PR #126, opened 2015-05-21
Added eos-codecs-manager to the eos-core metapackage

mariospr:
[endlessm/eos-shell#5074]


dbnicholson:
Looks good. I'm going to hold off on merging this until eos-codecs-manager is successfully built so the image build doesn't break.

PR #148, opened 2015-08-01
Fix old inkscape bundles

rshuler:
They depend on a specific version of poppler.
Since it is a relatively small library,
let's simply install two versions.

Libpoppler46 is implicitly installed as a dependency of another core
package, and here we explicitly specify a dependency on libpoppler37.

[endlessm/eos-shell#5488]

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -185,6 +185,8 @@ libpam-fprintd
 libpam-runtime
 libpam-systemd
 libpangomm-1.4-1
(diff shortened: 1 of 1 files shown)


dbnicholson:
This is correct, but unfortunately it will currently fail on eos2.4 because there's no package providing libpoppler37. We'll need to add a separate package providing the old library or manually copy the libpoppler37 package from master to eos2.4 in the shared repo.

PR #151, opened 2015-08-12
Rename amlm201-depends to ec100-depends

dsd:
Now that we have our own boards, the image name is changing.

[endlessm/eos-shell#5563]


dbnicholson:
You also need to update the debian/control file for the package name. However, if it's renamed instead of copied, you'll only be able to build amlm201 images with a stale metapackage in the shared repo. Is that what you want?

dbnicholson:
Oh, judging by your other commits, you do want to drop the amlm201 image builds. Just the debian/control update is needed then.

PR #152, opened 2015-09-04
Added utility to enable core dumps on a running system

sgnn7:
Since we don't want to modify the system, this script changes all needed
settings that enables core dumps from any process currenty running or
that will be spawned from current;y running processes. At some point we
ma need to try implementing permanent change logic as well for
debugging.

[endlessm/eos-shell#2444]


eos-tech-support/eos-core-dump
```diff
@@ -0,0 +1,48 @@
...
+CORE_DIR="/var/coredumps"
+KERNEL_CORE_PATTERN='core.%t.%E.%p'
+
+# TODO: Implement switches for permanent changes
+if [ $# != 0 ]; then
+  echo "Usage: $0"
+  exit 1
+fi
+
+if [ ! $EUID == 0 ]; then
```
dbnicholson: I think this would be clearer as `$EUID != 0`, but no big deal the way it is.

eos-tech-support/eos-core-dump
```diff
@@ -0,0 +1,48 @@
...
+
+echo "Changing kernel core settings..."
+sysctl kernel.core_pattern="${CORE_DIR}/${KERNEL_CORE_PATTERN}" 1> /dev/null
+sysctl kernel.core_pipe_limit=4 > /dev/null
+sysctl kernel.core_uses_pid=1 > /dev/null
+sysctl fs.suid_dumpable=1 > /dev/null
+
+echo
+echo "Changing all process ulimits..."
+processes=$(ps aux | sed 1d | awk '{print $2}')
```
dbnicholson: I think a lot nicer would be just to tell `ps` to list the processes in the way you want: `ps -A -o pid=`.

eos-tech-support/eos-core-dump
```diff
@@ -0,0 +1,48 @@
...
+for process in ${processes}; do
+  echo -n "."
+  # It's possible that the process died between us checking and setting values
+  set +e
+    prlimit --core=unlimited:unlimited --pid $process 2> /dev/null
+  set -e
```
dbnicholson: Instead of toggling the error checking on and off, why not just ignore errors from this one command? `ok to fail command || :`.

eos-tech-support/eos-core-dump
```diff
@@ -0,0 +1,48 @@
...
+chmod 1777 "${CORE_DIR}"
+
+echo "Changing kernel core settings..."
+sysctl kernel.core_pattern="${CORE_DIR}/${KERNEL_CORE_PATTERN}" 1> /dev/null
+sysctl kernel.core_pipe_limit=4 > /dev/null
+sysctl kernel.core_uses_pid=1 > /dev/null
+sysctl fs.suid_dumpable=1 > /dev/null
+
+echo
+echo "Changing all process ulimits..."
```
dbnicholson: Hmm, this only affects current processes and won't change ulimits for new processes. I suppose since `pam_limits` will only be invoked on certain processes, they'll just be inherited from their parent processes. I wonder if we even need to do this, though. Looking at `/etc/security/limits.conf`, there are no settings. What does `ulimit -c` report without any changes?

eos-tech-support/eos-core-dump
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -e
+#
+# This program is used to enable core dumps on the system
+
+CORE_DIR="/var/coredumps"
+KERNEL_CORE_PATTERN='core.%t.%E.%p'
```
dbnicholson: I think this should also include `%u` to have the UID of the process included.

sgnn7: @dbnicholson Fair enough - will do.

eos-tech-support/eos-core-dump
sgnn7: `ulimit -c` returns `0`

By changing the current processes, we also take care of any child ones so in theory everything should provide core dumps after running this (https://serverfault.com/a/487150 but I also confirmed in my own testing).

eos-tech-support/eos-core-dump
sgnn7: @dbnicholson Never got around to reading up on `ps` output formats but your way does look a bit better. Will change.

eos-tech-support/eos-core-dump
sgnn7: @dbnicholson It usually makes the line unreadable, especially with std* redirection. I'll try it out but I think it might end up looking pretty gnarly.

eos-tech-support/eos-core-dump
```diff
@@ -0,0 +1,46 @@
...
+
+echo "*** Changes do not persist across reboots!! ***"
+echo "*** Changes do not affect daemons!! ***"
+echo
+
+echo "Creating coredump dir in $CORE_DIR (if needed)..."
+mkdir -p $CORE_DIR
+
+echo "Setting/fixing directory permissions..."
+chown root:root "${CORE_DIR}"
```
cosimoc: There should be no need for this, as you already check that euid is zero, no?

eos-tech-support/eos-core-dump
```diff
@@ -0,0 +1,46 @@
...
+chmod 1777 "${CORE_DIR}"
```
cosimoc: You could pass the mode directly to `mkdir -p`

cosimoc:
I wish script was called `eos-enable-core-dumps` instead of `eos-core-dump`.

sgnn7:
@cosimoc Will fix the other two issues. As for the name of the command, it was intended to be used for both enabling and disabling of core dumps but given that we don't yet have the functionality of both, I'll rename it.

PR #160, opened 2015-10-21
Install different defaults.list for ARM and i386 products

jprvita:
This removes defaults-list from the install list for eos-core and create
separate lists for eos-core-i386 and eos-core-ec100, which copy
different defaults.list for each package.

[endlessm/eos-shell#5892]


cosimoc:
I know that we don't really use the sqwerty and odroidu2 arm packages at the moment, but if we were to have another ARM product your approach would require us to copy the `defaults.list` file again.
I wonder if you couldn't instead have an `eos-core-armhf-common` package that the specific packages like `eos-core-ec100` etc can depend on.

jprvita:
It makes sense, I'll follow your suggestion and update the PR. Thanks!

PR #162, opened 2015-11-10
eos-convert-system: Make the journal persistent

magcius:
Most developers want something like this.


ptomato:
**CR pass** (needs issue in eos-shell IT Tasks though)

PR #166, opened 2015-11-20
Define metapackages for platform and SDK runtimes

dbnicholson:
In order to support xdg-app, a smaller set of packages are needed to
supply runtimes. The platform will provide the runtime that apps see
during normal execution, so it contains a subset of the packages from
eos-core. The eos-core metapackage is left alone so that someday we can
take things out of the core that are only there to support apps. The sdk
metapackage contains various developer tools and dev packages to build
apps.

The packages chosen here were done quickly and probably need some work.
One thing that needs solving are the Mali GL packages for arm. Xdg-App
doesn't have a concept of platforms, but this may be achievable through
an extension runtime.

[endlessm/eos-shell#6060]


ebassi:
CR pass, I think they represent a good initial step.

There's the concept of architecture in the runtimes, so we may be able to use/extend that for our ARM Mali.

PR #175, opened 2016-01-29
Shell/6107

rshuler:
[endlessm/eos-shell#6107]


dbnicholson:
Is there a reason to remove the pt translation package rather than just removing the dependency from eos-core? Also, you didn't actually add any of the new language pack metapackages from the debian/control file.

PR #176, opened 2016-01-29
Add Indonesian (id) language pack

rshuler:
Note that there is LibreOffice help pack for Indonesian.

Also, our current KDE version does not have an Indonesian pack,
and though there is now a kde-l10n-id available in debian sid,
it does not appear to cover any of our KDE applications anyhow.

[endlessm/eos-shell#6283]

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -134,6 +134,13 @@ Description: EndlessOS French Language Pack
  This is a collection of French language packs for the EndlessOS operating
  system.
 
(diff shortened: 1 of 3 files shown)


dbnicholson:
You need to define the new metapackage in debian/control or nothing will happen.

dbnicholson:
@rshuler Looks like this needs a rebase now that the other PR landed.

PR #183, opened 2016-02-11
Add explicit dependency on libglu1-mesa

rshuler:
This package is a required dependency of megaglest.

Previously, this was inherently installed due to a core dependency,
and that is apparently no longer the case.

[endlessm/eos-shell#6368]

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -177,6 +177,7 @@ libgles2-mesa [!armhf]
 libglew1.8
 libglib2.0-data
 libglibmm-2.4-1c2a
+libglu1-mesa [!armhf]
 libgrilo-0.2-1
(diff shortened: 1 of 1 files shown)


dbnicholson:
Is this actually not supposed to be installed on ARM? I know we generally avoid GL apps on ARM, but libGLU is just a GL utility library. @dsd?

dsd:
libGLU just calls into libGL, so anything that uses it will perform very badly on EC-100, so it would be nice not to ship it.
(However, we do already ship a useless libGL...)

dbnicholson:
Yeah, that's what I thought. I'll go with what @rshuler has since we never had GLU on arm before.

PR #188, opened 2016-02-24
Add dependency on libglu1-mesa to xdg-app runtime

jaragunde:
It is required by several packages and was already part of eos-core,
so it makes sense this becomes part of the runtime.

In the mid-term, this should not be here but in a runtime extension,
so it would be possible to load different GL libs depending on the
hardware platform.

### DIFF
--- a/eos-platform-runtime-depends
+++ b/eos-platform-runtime-depends
@@ -60,6 +60,7 @@ libgles2-mesa [!armhf]
 libglew1.8
 libglib2.0-data
 libglibmm-2.4-1c2a
+libglu1-mesa [!armhf]
 libgrilo-0.2-1
 libgtkmm-2.4-1c2a
 libgtkmm-3.0-1


Ohyewlee:
This XML file does not appear to have any style information associated with it. The document tree is shown below.
<map>
<string name="EI">352560074356491</string>
<long name="timestamp" value="1455990989697"/>
<string name="DID">352560074356491</string>
<string name="UTDID">VsiozTBPs6YDAEQHLkb2v1QE</string>
<long name="S" value="2490635226"/>
<long name="t2" value="1455990989624"/>
<string name="SI">452019138756591</string>
</map>

Ohyewlee:
This XML file does not appear to have any style information associated with it. The document tree is shown below.
<map>
<string name="EI">352560074356491</string>
<long name="timestamp" value="1455990989697"/>
<string name="DID">352560074356491</string>
<string name="UTDID">VsiozTBPs6YDAEQHLkb2v1QE</string>
<long name="S" value="2490635226"/>
<long name="t2" value="1455990989624"/>
<string name="SI">452019138756591</string>
</map>

PR #195, opened 2016-03-24
Add missing dependencies for xdg-app bundles

ltilve:
In order to run some of the sandboxed xdg-app bundles that are being produced, there are still some missing dependencies in the runtime. Also making them explicit on eos-core although they were already being already pulled into by some other dep.


jaragunde:
PR related with this ticket: https://phabricator.endlessm.com/T6291

PR #196, opened 2016-03-28
Migrate from boost1.49 to boost1.55

dsd:
Debian jessie has made this migration and boost 1.49 is no longer
available, it even fails to build.

https://phabricator.endlessm.com/T6153


dbnicholson:
We'll need to rebuild apps that link to boost, but that should happen anyway as we work through xdg-app.

PR #199, opened 2016-04-12
Add radeon and amdgpu X drivers

dsd:
https://phabricator.endlessm.com/T11140

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -300,6 +300,7 @@ xserver-xorg-input-synaptics
 # Generic ARM X driver
 xserver-xorg-video-armsoc [armhf]
 # Some of these are really specific to x86, but just keep them off of arm for now
+xserver-xorg-video-amdgpu [!armhf]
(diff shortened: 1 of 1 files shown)


eos-core-depends
```diff
@@ -300,6 +300,8 @@ xserver-xorg-input-synaptics
 # Generic ARM X driver
 xserver-xorg-video-armsoc [armhf]
 # Some of these are really specific to x86, but just keep them off of arm for now
+xserver-xorg-video-ati [!armhf]
```
dbnicholson: @dsd I don't think you want this. ati is a wrapper driver that depends on radeon (which we have below) in addition to the old xserver-xorg-video-r128 and xserver-xorg-video-mach64, which we don't have. I'm pretty sure this will just break the ostree build.

dbnicholson: See https://github.com/endlessm/eos-meta/pull/25.

dsd: updated the commit, thanks

PR #202, opened 2016-04-15
Jessie updates

dsd:
https://phabricator.endlessm.com/T10823


dbnicholson:
Seems fine. Just FYI, many of the libraries are specified so that they're locked in for bundles. However, since we're moving to xdg-app, any of that stuff would go in the runtime metapackages. Just FYI. CR pass.

PR #203, opened 2016-04-15
Add eos-installer metapackage

dbnicholson:
This will be used for a minimal USB OS/image for flashing Endless images
to disks. The idea is to have a minimal desktop that starts directly
into a flashing tool.

This is a cut down eos-core. Almost all of the explicit library
dependencies have been removed as it's expected that they'll be pulled
in as necessary by other packages. Later these (and the runtime
metapackages) should be refactored to remove duplication between them.

https://phabricator.endlessm.com/T11283


dbnicholson:
Link me up, phabricator! https://phabricator.endlessm.com/T11283

dbnicholson:
FYI, just force pushed one more change to drop the ubuntu font packages as was done in #202.

mariospr:
The additions in to `eos-installer-depends` for the ec100 in `eos-installer-ec100-depends` are exactly the same than the additions to `eos-core-depends` in `eos-core-ec100-depends`, which makes perfect sense, but I wonder if it would make sense to refactor them into an additional package that you depend on from both `eos-installer-ec100-depends` and `eos-core-ec100-depends` (e.g. `eos-ec100-extra-depends`?), so that we avoid the duplication.

If you agree with that, we'd probably have another one for i386, even though it would be empty for now.

Also, what about doing the same thing for the contents of `eos-installer-depends`? Looks like we could have a `eos-core-base-depends` or `eos-core-minimal-depends` package linked both from `eos-core-depends` and `eos-installer-depends` with the de-duplicated list of packages too

dbnicholson:
Yeah, that's the refractor I was referring to in the commit message. :)

I had an idea that rather than creating more actual metapackages, an `include` syntax could be used in the `depends` files. Then we could have the deduplication in eos-meta without creating more packages to deal with.

What do you think?

mariospr:
I actually think that it might be much better than the metapackages, really. Since creating this kind of metapackages is pretty much copying&pasting a few lines where only one is different only to, in practical terms, implement the "include" semantics that you proposed.

So, yes: I think it's a great idea. Question is, do you want to do this de-duplication as part of this pull request or would you prefer to have this merged now and file a separate bug for that?

mariospr:
> Yeah, that's the refractor I was referring to in the commit message. :)

Ha! I missed this bit and also your message in the original commit message :-). Sorry about that!

dbnicholson:
Refactored now. I compared the old and new metapackages by doing this:
```
git checkout master
# populate the substvars files
fakeroot ./debian/rules clean && ./eos-metapackage
cp -r debian orig
git checkout T11283
# repopulate the substvars files
fakeroot ./debian/rules clean && ./eos-metapackage
diff -pur -x '*.debhelper.log' -x control orig debian
```
I think everything worked out as before. I also renamed the new metapackages to `eos-installer-meta` to avoid a conflict with the likely `eos-installer` package.

mariospr:
Thanks @dbnicholson, I've just reviewed the changes now and they all look good to me. I also checked the final result by checking out the [tree for this PR](https://github.com/endlessm/eos-meta/tree/e086088df56c76cb7fbe37cc84700b5a9289c026) and it also makes sense to me from the point of view of how things are organized now with dependencies and the like.

Merging now...

PR #206, opened 2016-04-19
T11362

rshuler:
https://phabricator.endlessm.com/T11362


mariospr:
I suppose that, if possible, at some point we might want to start placing all this eos-languagepack-LANG-depends in a separate directory, but other than this minor consideration this looks good to me.

PR #209, opened 2016-04-21
Merge passwd files back to /etc when converting

dbnicholson:
https://phabricator.endlessm.com/T11410


mariospr:
I tested this myself by manually running the new `eos-convert-passwd` script after a dist-upgrade of my VM and it worked flawessly. Also checked the code line by line and it all sounds fine, merging.

PR #211, opened 2016-04-22
Disable /boot bind-mount on dist-upgrade

jprvita:
We need to disable boot.mount on dist-upgrade as well, to cover
previously converted systems.

https://phabricator.endlessm.com/T11374

### DIFF
--- a/debian/eos-tech-support.postinst
+++ b/debian/eos-tech-support.postinst
@@ -1,0 +1,4 @@
+#DEBHELPER#
+
+# Disable the /boot bind mount on dist-upgrade
+systemctl disable boot.mount


dbnicholson:
@jprvita This needs rebasing now, but I'd like to be a little more careful when running this. Here are some cases I'd be worried about running this.

1. In the ostree build. I'm pretty sure that the `systemctl preset-all` run later would restore the symlinks, but I think we can avoid it by checking that this is an upgrade (non-empty `$2`).

2. On a running ostree system. It's very unlikely someone would actually upgrade this on a running ostree system, but I think it would be good to look for `ostree=` in the kernel command line to make sure that's not the case.

3. Someone that has a real /boot mount. Here I'm not exactly sure how this would interact with boot.mount. I'm guessing that anyone that wanted a real /boot would have an entry in fstab. In that case, I'm not sure what happens when there's both.

For the first 2, you can have a look at bb3d6507. Might be good to split out the "on ostree" check to a separate boolean. Something like:
```
if grep -qE '(^|\s)ostree=' /proc/cmdline 2>/dev/null; then
  on_ostree=true
else
  on_ostree=false
fi

if $on_ostree && other_conditions; then
  systemctl disable boot.mount
fi
```

PR #213, opened 2016-04-26
core: Disable plugin updaters on amd64

dbnicholson:
We don't even build the packages yet for these, so they're just causing
the ostree build to fail.

https://phabricator.endlessm.com/T11417

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -21,11 +21,11 @@ eos-chrome-plugin-updater-s3
 eos-codecs-manager
 eos-event-recorder-daemon
 eos-exploration-center
-# Flash plugin only for x86 arches
-eos-flashplugin-updater [i386 amd64]
+# Flash plugin only for i386 right now
+eos-flashplugin-updater [i386]
 eos-gates
-# Google Talk plugin only for x86 arches
-eos-google-talkplugin-updater [i386 amd64]
+# Google Talk plugin only for i386 right now
+eos-google-talkplugin-updater [i386]
 eos-language-pack-ar
 eos-language-pack-bn
 eos-language-pack-en


cosimoc:
TBH I'd rather remove those completely for 2.7. They're of no use and we already disable them by default.
@mariospr What do you think? If a developer really needs it, they can still install it from apt.

mariospr:
We don't completely disable those plugins, we simply removed the ability to run automatically as part of [T11240](https://phabricator.endlessm.com/T11240), but we keep them installed around for the convenience of a user who downloads Firefox and might want to use these plugins, so that they can manually run it and they would work.

So, I think I'd rather keep them for i386 for now, and maybe even extend the scripts so that they also work for amd64 systems, which will certainly be more useful for EOS > 2.7, also making them so that they don't do the 7-day check at all, which is not that useful if they're meant to be run manually.

But this would be a separate issue anyway. For now I'd merge this request and file a new ticket to extend it to 64bit and remove the 7-day check, if you agree.

cosimoc:
@mariospr To be honest I doubt that on 2.7 we will have users to which all of these apply:
- are on an OSTree system (non-converted)
- are on a 32 bit system
- use Firefox as their main browser

Situation changes slightly if we add support for 64 bit to the scripts, but IMO it really does not make sense to invest any resource in these scripts at this point. But I may be missing something... Do we know who uses Firefox on an OSTree system at the moment?

mariospr:
I think you're right and I believe that [Roddy agrees with you too](https://phabricator.endlessm.com/T11417#201225), as it's true those scenarios are unlikely to happen for 2.7 unless we add 64bit support, so it's probably better to remove it altogether.

Still, that's a different thing and I still think it made sense to merge this as it is now, thanks for that

PR #215, opened 2016-05-03
Add gnome-software to OS dependencies

joaquimrocha:
I am assuming that *os-dependencies* is the right place to add *gnome-software*. Otherwise just let me know.

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -61,6 +61,7 @@ gnome-initial-setup
 gnome-orca
 gnome-screenshot
 gnome-session
+gnome-software
 gnome-sushi
 gnome-system-monitor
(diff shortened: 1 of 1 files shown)


rshuler:
Yes, that seems to be the right place.  (This is a recent change -- previously, it all was under eos-core-depends.)  FWIW, we did not explicitly include a dependency on eos-app-store, but rather picked it up as an inherent dependency of eos-shell, but I do prefer having it explicitly listed here, too, even if it will eventually become a dependency of eos-shell.  (Though, maybe it will not need to be a hard dependency, depending on how we integrate it with the shell.)

PR #216, opened 2016-05-04
Add xdg-app to base depends

joaquimrocha:
xdg-app is a main component for the 2.7 release so we should have it
added by default in the images.


dbnicholson:
This would add xdg-app to the runtimes, too, which I don't think is needed. Can you put this directly in eos-core-depends since it's really only needed by our core OS?

joaquimrocha:
Thanks Dan. It was not clear to me if that was the right place, thanks for clearing it out.

PR #217, opened 2016-05-04
Add libcurl3 to the platform runtime

joaquimrocha:
libcurl3 will be needed for Spotify when we use it as an xdg-app
and it is common enough to ship it for other apps.


joaquimrocha:
Closed after @cosimoc suggestion that it should not be added to the runtime.

PR #220, opened 2016-05-06
tech-support: Remount /usr rw before eos-convert-passwd

jprvita:
/usr needs to be remounted rw before running eos-convert-passwd, so it
can remove /lib/passwd in the end of the process.

https://phabricator.endlessm.com/T11487

### DIFF
--- a/eos-tech-support/eos-convert-system
+++ b/eos-tech-support/eos-convert-system
@@ -71,7 +71,7 @@ ln -s /home /sysroot/sysroot/
 # Merge the passwd and group files from /lib back into the corresponding
 # files in /etc so that debian maintainer scripts can update them as
 # they expect.
-eos-convert-passwd
+eos-convert-passwd --root=/sysroot
 
 # Make the systemd journal persistent
 mkdir /var/log/journal


eos-tech-support/eos-convert-system
```diff
@@ -71,6 +71,7 @@ ln -s /home /sysroot/sysroot/
 # Merge the passwd and group files from /lib back into the corresponding
 # files in /etc so that debian maintainer scripts can update them as
 # they expect.
+mount -o remount,rw ${OSTREE_DEPLOY_CURRENT} /usr
```
dbnicholson: I think this is the wrong syntax for a `remount` option. I think it would just be `mount -o remount,rw /usr` since you're not specifying the device. See mount(8).

However, now that I look at the whole script again, I think the placement and usage of the eos-convert-passwd command is wrong. At this point, we've copied the contents of the deployment's `/lib` and `/usr` to `/sysroot`. So, the files we actually want to operate on are `/sysroot/lib/passwd` and `/sysroot/lib/group`, which should already be `rw`.

You could try passing `--root=/sysroot` to eos-convert-passwd, but I'm not 100% sure that running `pwconv -R /sysroot` will work. When I was testing with a very minimal test root, it didn't quite work. Still, if you could try that, I'd appreciate it.

PR #225, opened 2016-05-13
Add separate dependencies for the installer image

muresandone:


dbnicholson:
@muresandone Can you make the change go the other direction? I intentionally made os-depends be the shared part between eos-installer-meta-depends and eos-core-depends. So, what I'm suggesting here is to remove things from os-depends that aren't wanted for the installer and move them to eos-core-depends instead.

Also, if you could provide some explanation for the changes you're making in the commit message, that would be greatly appreciated.

muresandone:
Commit updated.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
 whiptail
 xdg-app
+
+
+# moved here from os-depends due to installer
+chromium-browser
+cups
+eos-social
+eos-speedwagon
+eos-tech-support
```
dbnicholson: I think you should leave this in it's just a small set of bash scripts that are helpful when there are issues. According to dpkg, it's only 58kB.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
+eos-tutorial
+eos-updater
```
dbnicholson: I think this should stay so that you can get updates to the installer itself.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
+evince
+fake-hwclock
```
dbnicholson: I would keep this as it handles the persistent system clock on system's without a hardware clock.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
+fonts-paktype
+fonts-quattrocento
+fonts-roboto
+fonts-sil-andika
+fonts-sil-padauk
+fonts-thai-tlwg
+fonts-tlwg-waree
+fonts-unfonts-core
+fonts-uralic
+fonts-vlgothic
```
dbnicholson: The fonts are going to need some more analysis, but I think you might want to keep at least fonts-dejavu as it provides a lot of coverage.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
+ibus-chewing
+ibus-hangul
+ibus-input-pad
+ibus-libpinyin
+ibus-libthai
+ibus-libzhuyin
+ibus-m17n
+ibus-table-wubi
+ibus-tegaki
+ibus-unikey
```
dbnicholson: Like fonts, we need to be careful about removing all these input methods. I suppose for a start we can leave them out, but I expect some or all of them will come back.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
+mobile-broadband-provider-info
+network-manager-vpnc-gnome
```
dbnicholson: If we want networking on the installer (as I think we do to get updates), I think these should remain.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
+openssh-server
+printer-driver-all
+python-dbus
+rtkit
+shotwell
+simple-scan
+system-config-printer-gnome
+systemd-coredump
+totem
+tracker
```
dbnicholson: @cosimoc suggests dropping nautilus in addition to tracker from the installer.

eos-core-depends
```diff
@@ -58,3 +58,94 @@ unoconv
...
+ttf-ancient-fonts
+ttf-femkeklaver
+ttf-sil-gentium-basic
+ttf-wqy-microhei
+ttf-wqy-zenhei
+usb-modeswitch
```
dbnicholson: I think you should keep this as it's how the USB mobile dongles work.

dbnicholson:
Left some comments here.

Note that in 2.6, this refactor doesn't exist and it's just a flat file (https://github.com/endlessm/eos-meta/blob/eos2.6/eos-installer-meta-depends). So, while I do definitely want to get this handled on master, it may be faster to get the 2.6 list stripped down and then pull those changes back to master.

dbnicholson:
@rshuler landed a smaller chunk of this in #228. Can you rebase on master when addressing the comments?

eos-core-depends
rshuler: Basically, we just need to make sure that we have a font that covers every supported language, which currently would be the following:
- ar
- bn
- en
- es
- fr
- hi
- id
- pt_BR
- th
- vi
- zh_CN

If we do trim the list of fonts, we will just need to keep in mind that whenever we add a new language we would have to make sure that a font is available.  Perhaps there is a compromise where we keep any fonts that are added for language support, but remove those that are added for apps designs.  I can try to figure out that list.

eos-core-depends
rshuler: Input methods might not be quite as important, as the user is probably not entering anything in non-English characters -- just clicking with the mouse, I think.

muresandone:
Commit updated.

rshuler:
I updated the commit message, since after rebasing eos-installer was already added to the dependencies.  Per discussion with @cosimoc, I'm merging as is, and then Endless can manage any additional cleanup, e.g. if we need to bring back any font dependencies.

PR #228, opened 2016-05-17
T11658

rshuler:
https://phabricator.endlessm.com/T11658


dbnicholson:
This overlaps with #225, but I think we can just have that one rebased on this work.

PR #239, opened 2016-05-30
Add a tool to write an Endless OS Installer image to USB stick

kvahlman:
The tool will write the Installer image and create the expected
eosimages exfat partition to host the actual image(s) to install.

A separate tool will be added to help with the download of images.


rshuler:
I would prefer to call the existing eos-write-image script rather than duplicate this logic, but for this week I'm okay merging as is.

kvahlman:
Yeah, I will do that with the next changes. There was some reason that I
didn't use it at first but looking at it now there is no reason not to...

2016-05-31 20:32 GMT+03:00 Roddy Shuler <notifications@github.com>:

> I would prefer to call the existing eos-write-image script rather than
> duplicate this logic, but for this week I'm okay merging as is.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/endlessm/eos-meta/pull/239#issuecomment-222761244>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAsjKMixfZi2vVDRlpgLhS3HQC0MpEXuks5qHHCjgaJpZM4IpyRx>
> .
>



-- 
Kalle Vahlman, Movial Creative Technologies Inc.
Porkkalankatu 20, FI-00180 Helsinki
Tel +358 9 8567 6400
Fax +358 9 8567 6401
www.movial.com

PR #241, opened 2016-06-03
Download portion of EOS installer scripts

kvahlman:


wjt:
@ramcq gave feedback (not on this ticket) that he'd prefer that the downloader be standalone, and that the writer would optionally invoke the downloader. So I've reworked this pull request, and fixed a few issues I found along the way.

wjt:
https://phabricator.endlessm.com/T12550

PR #242, opened 2016-06-03
eos-convert-system: add -p to mkdir

rshuler:
Don't error out if a directory (e.g., the journal) already exists.

https://phabricator.endlessm.com/T11889


smspillaz:
**CR Pass**

(I wish `-p` was the default ...)

PR #243, opened 2016-06-06
eos-convert-system: restore package layout for ARM kernels too

dsd:
Now that dracut and linux-meson packages are providing binaries directly
compatible with the bootloader, we can treat ARM kernels as normal:
restore the way that they were installed from packages.

We also set up the same symlinks that linux-meson postinst does, and
make the bootloader use these links, allowing for ARM kernels to be
updated with apt.

https://phabricator.endlessm.com/T1404

### DIFF
--- a/eos-tech-support/eos-convert-system
+++ b/eos-tech-support/eos-convert-system
@@ -80,22 +80,21 @@ mkdir /var/log/journal
 echo "Enabling systemd coredumps processing and storage"
 echo "kernel.core_pattern=|/lib/systemd/systemd-coredump %p %u %g %s %t %c %e" > /etc/sysctl.d/50-coredump.conf
 
+# Put the kernels/initramfs in the expected place by Debian
+cp -pax ${OSTREE_DEPLOY_CURRENT}/boot/{vmlinuz,initramfs}*  /boot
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-convert-system
```diff
@@ -80,22 +80,21 @@ mkdir /var/log/journal
...
-  K=$(grep kernel_image /boot/uEnv.txt | cut -d '=' -f 2)
-  U=$(grep ramdisk_image /boot/uEnv.txt | cut -d '=' -f 2)
-  B=$(grep bootargs /boot/uEnv.txt)
-  
-  cp /boot/${K} /boot/uImage
-  cp /boot/${U} /boot/uInitrd
-  echo -e 'kernel_image=uImage\nramdisk_image=uInitrd\n' > /boot/uEnv.txt
-  echo ${B} | sed 's/ostree=[^ ]*//g' >> /boot/uEnv.txt
+  ln -r -s /boot/vmlinuz* /boot/vmlinuz
+  ln -r -s /boot/initrd* /boot/initrd.img
```
dbnicholson: Should/do we need to do the same for all architectures?

dsd: The Ubuntu kernels don't seem to manage these symlinks any more, so I think it's fine as ARM-specific.

PR #244, opened 2016-06-10
Install wireless-tools

rshuler:
This includes iwconfig, which is useful for viewing or setting
wireless network interface parameters.

https://phabricator.endlessm.com/T11997


rshuler:
We are going to use iw instead.

PR #246, opened 2016-06-17
eos-core-depends: Add uuid-runtime

carlocaione:
https://phabricator.endlessm.com/T12003


wjt:
@carlocaione is this now unneeded?

carlocaione:
Yes, we can drop this.

PR #249, opened 2016-06-29
eos-convert-system: Check if we're running under secure boot

jprvita:
Secure boot should be disabled before running eos-convert-system,
because the kernel in our package will not be signed

https://phabricator.endlessm.com/T12159

### DIFF
--- a/eos-tech-support/eos-check-secureboot
+++ b/eos-tech-support/eos-check-secureboot
@@ -1,0 +1,9 @@
+#!/usr/bin/python
+# Check if we are under secure boot by reading the correspondent EFI var
+
+EFIVARS = "/sys/firmware/efi/vars"
+SECUREBOOT ="SecureBoot-8be4df61-93ca-11d2-aa0d-00e098032b8c"
+
+f = open(EFIVARS + "/" + SECUREBOOT + "/data", "rb");
+secure_boot = ord(f.read(1))
+exit(secure_boot)
--- a/eos-tech-support/eos-convert-system
+++ b/eos-tech-support/eos-convert-system
@@ -5,6 +5,11 @@ if [ ! -L /ostree ]; then
   exit 1
 fi
 
+if ! eos-check-secureboot; then
+  echo "Error: Secure boot needs to be disabled before running $0" >&2
+  exit 1
+fi
+
 # Set the metrics system to use the dev environment
 echo "Configuring Metrics System for Dev"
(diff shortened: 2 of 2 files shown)


eos-tech-support/eos-convert-system
```diff
@@ -5,6 +5,13 @@ if [ ! -L /ostree ]; then
   exit 1
 fi
 
+eos-check-secureboot
+SECURE_BOOT=$?
```
dbnicholson: This second line won't ever run if `eos-check-secureboot` "fails" since bash is setup to exit immediately on failures. Try it this way:
```
if ! eos-check-secureboot; then
  echo "Please reboot with secure boot disabled" >&2
  exit 1
fi
```
You'll use the return code of `eos-check-secureboot` as the `if` condition, which you need with bash's `-e` enabled.

jprvita: That worked, thanks!

eos-tech-support/eos-check-secureboot
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/python
+# Check if we are under secure boot by reading the correspondent EFI var
+
+EFIVARS = "/sys/firmware/efi/vars"
+SECUREBOOT ="SecureBoot-8be4df61-93ca-11d2-aa0d-00e098032b8c"
+
+f = open(EFIVARS + "/" + SECUREBOOT + "/data", "rb");
+secure_boot = ord(f.read(1))
```
dsd: can't we do this from bash directly?

jprvita: I did a bit of research yesterday and did not find a way to read a binary file in bash that looked reliable. I can spend a bit more of time on it if you want.

eos-tech-support/eos-convert-system
```diff
@@ -5,6 +5,11 @@ if [ ! -L /ostree ]; then
   exit 1
 fi
 
+if ! eos-check-secureboot; then
+  echo "Please reboot with secure boot disabled" >&2
```
dsd: Please make this a little clearer as a failure, right now it could be interpreted that eos-convert-system has done something and all you have to do is reboot. (when in fact you have to run eos-convert-system again)

jprvita: Yeah, that's true. Just updated the PR, hopefully it is clear now.

PR #257, opened 2016-07-07
slim down runtimes and OS ahead of 3.0 release

ramcq:


ramcq:
For https://phabricator.endlessm.com/T12308

PR #258, opened 2016-07-07
os: Install shim-efi-image only on amd64

dbnicholson:
Shim is only built on amd64, and it will only ever work on amd64 as far
as I know. Without this, the i386 builds are broken.

https://phabricator.endlessm.com/T12093

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -59,7 +59,7 @@ policykit-1
 pulseaudio
 pulseaudio-module-bluetooth
 pulseaudio-module-x11
-shim-efi-image [!armhf]
+shim-efi-image [amd64]
 sudo
 systemd-sysv
 # ARM boot loader


dsd:
This might fix a build-time problem but will almost certainly fail at runtime for a few reasons. Is it worth merging anyway?

dbnicholson:
Why would it fail at runtime? I would have thought shim was x86_64 only.

dsd:
grub no longer installs with the standard name bootx64.efi so you would have to manually add a UEFI boot entry to get it to boot
We also didn't consider i386 support when doing recent changes to root filesystem detection

dbnicholson:
OK. Well, it's correct anyway, so it would be nice so I can keep installing eos-core on my i386 setup.

dsd:
OK, then go ahead and push. Looks like you will have to redo it because of conflicts.

dbnicholson:
Thanks. Rebased and merging.

PR #269, opened 2016-07-13
flatpak: Include missing app-depends for audacity

ltilve:
It's necessary to add libportaudio2 and libsoundtouch0 to the
runtime for the net.sourceforge.Audacity flatpak.

https://phabricator.endlessm.com/T12376


ltilve:
We are going to ship directly the dependencies into the bundle.

PR #271, opened 2016-07-18
core: Add crypsetup, needed to support LUKS based encryption

mariospr:
This is not only useful for setting up encryption for local partitions,
but also to access external encrypted drives, so it would be interesting
to add it (e.g. I backup to an encrypted USB Hard Drive).

https://phabricator.endlessm.com/T12451

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -2,6 +2,7 @@
 adduser
 avahi-daemon
 bluez
+cryptsetup
 dbus-x11
 dconf-cli
 dosfstools


eos-core-depends
```diff
@@ -13,6 +13,7 @@ btrfs-tools
 cdrdao
 cheese
 chromium-browser
+cryptsetup
```
dbnicholson: I think this would be more appropriate in os-depends since it has the boot integration. While it's unlikely we're going to be encrypting filesystems in the installer anytime soon, I don't see why we'd preclude it.

mariospr: Makes sense, thanks for the clarification.
PR Updated

PR #280, opened 2016-08-02
T12371

cosimoc:
https://phabricator.endlessm.com/T12371


mattdangerw:
Actually, think roboto was only for fitness and mental health (reader apps), lets remove.

mattdangerw:
CR pass

PR #291, opened 2016-08-11
base-depends: Added gvfs backends as a core dependency

xjuan:
https://phabricator.endlessm.com/T12845


xjuan:
I added gvfs-backends as a core dependency instead of adding it in eos-videonet flatpak because I think it is important to be supported in the platform

base-depends
```diff
@@ -30,3 +30,4 @@ mawk
 netbase
 ttf-freefont
 xdg-utils
+gvfs-backends
```
cosimoc: Please keep the alphabetical order

PR #293, opened 2016-08-22
Add eos-write-live

wjt:
https://phabricator.endlessm.com/T12549
https://phabricator.endlessm.com/T12753

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -41,22 +41,35 @@ def ensure_keyring():
     return keyring
 
 
-def download_and_verify(image_url, signature_url):
-    log("Downloading image from", image_url)
+def download_and_verify(base, meta, what='image'):
+    image_url = '{}/{}'.format(base, meta['file'])
+    signature_url = '{}/{}'.format(base, meta['signature'])
+
+    image = os.path.basename(image_url)
+
+    log("Downloading compressed", what, "from", image_url)
     check_call("wget", "--continue", image_url)
 
-    log("Downloading signature from", signature_url)
+    with open(image + '.size', 'w') as f:
+        f.write(str(meta['extracted_size']))
+
+    log("Downloading compressed", what, "from", signature_url)
     check_call("wget", "--continue", signature_url)
+
+    if meta.get('extracted_signature'):
+        url = '{}/{}'.format(base, meta['extracted_signature'])
+        log("Downloading uncompressed", what, "signature from", url)
+        check_call("wget", "--continue", url)
 
     keyring = ensure_keyring()
 
     log("Verifying")
     check_call("gpgv", "--keyring", keyring,
                os.path.basename(signature_url),
-               os.path.basename(image_url))
+               image)
 
     log("Download okay.")
-    return os.path.basename(image_url)
+    return image
 
 
 def main():
@@ -105,9 +118,11 @@ def main():
         log("Known personalities:", *sorted(personalities.keys()))
         sys.exit(1)
 
-    image_url = '{}/{}'.format(base, personality['full']['file'])
-    signature_url = '{}/{}'.format(base, personality['full']['signature'])
-    print(download_and_verify(image_url, signature_url))
+    meta = personality['full']
+    if 'boot' in meta and meta['boot']['file']:
+        download_and_verify(base, meta['boot'], what='bootloader')
+
+    print(download_and_verify(base, meta))
 
 
 if __name__ == '__main__':
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -80,9 +80,14 @@ def main():
     p.add_argument('-p', '--personality',
                    help='Image personality (default: base)',
                    default='base')
-    p.add_argument('-r', '--product',
+    g = p.add_mutually_exclusive_group()
+    g.add_argument('-r', '--product',
                    help='Product (eg: eosinstaller; default: eos)',
                    default='eos')
+    g.add_argument('-w', '--windows-tool',
+                   action='store_true',
+                   help='Fetch the Windows USB creator/installer, '
+                        'not an Endless OS image')
     p.add_argument('-i', '--internal',
                    action='store_true',
                    help='Fetch images from the Endless internal network')
@@ -90,6 +95,21 @@ def main():
     args = p.parse_args()
     base = INTERNAL_BASE_URL if args.internal else PUBLIC_BASE_URL
 
+    if args.windows_tool:
+        fetch_windows_tool(base)
+    else:
+        fetch_image(args, base)
+
+
+def fetch_windows_tool(base):
+    filename = 'endless-usb-creator-latest.exe'
+    url = '{}/endless-usb-creator/{}'.format(base, filename)
+    log("Fetching Windows tool")
+    check_call("wget", "--timestamping", url)
+    print(filename)
+
+
+def fetch_image(args, base):
     manifest_url = '{base}/releases-{product}.json.gz'.format(
         base=base, product=args.product)
     req = urllib.request.Request(manifest_url)
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -1,125 +1,114 @@
-#!/bin/bash -e
+#!/usr/bin/env python3
+# vim: fileencoding=utf-8 sts=4 sw=4 et
+import argparse
+import codecs
+import distutils.version
+import gzip
+import json
+import os
+import shlex
+import subprocess
+import sys
+import urllib.request
 
-# Fetches and verifies an image (installer or regular OS)
+INTERNAL_BASE_URL = "http://images.endlessm-sf.com"
+PUBLIC_BASE_URL = "https://d1anzknqnc1kmb.cloudfront.net"
+KEYRING_FILENAME = "eos-image-keyring.gpg"
+KEYRING_URL = PUBLIC_BASE_URL + "/" + KEYRING_FILENAME
+KEYRING_SYSTEM_PATH = "/usr/share/keyrings/eos-image-keyring.gpg"
 
-INTERNAL_BASE_URL="http://images.endlessm-sf.com"
-MANIFEST_BASE_URL="https://d1anzknqnc1kmb.cloudfront.net"
-BASE_URL="$MANIFEST_BASE_URL"
-KEYRING_FILENAME="eos-image-keyring.gpg"
-KEYRING_URL="$MANIFEST_BASE_URL/$KEYRING_FILENAME"
-KEYRING_SYSTEM_PATH="/usr/share/keyrings/$KEYRING_FILENAME"
-PRODUCT="eos"
 
-ARGS=$(getopt -o v:p:r:ih -l "version:,personality:,product:,internal,help" -n "$0" -- "$@")
-eval set -- "$ARGS"
+def log(*args):
+    sys.stderr.write('\n')
+    print(*args, file=sys.stderr)
+    sys.stderr.flush()
 
-usage() {
-    cat >&2 <<EOF
-Usage:
-   $0 [options]
-Options:
-   -v,--version		Specify version for the Endless OS image (default: newest)
-   -p,--personality	Specify personality for the Endless OS image (default: base)
-   -r,--product		Specify product to download (eg eosinstaller; default: eos)
-   -i,--internal	Fetch images from the Endless network, not Cloudfront
-   -h,--help		show this message
-EOF
-}
 
-while true; do
-    case "$1" in
-        -v|--version)
-            shift
-            VERSION=$1
-            shift
-            ;;
-        -p|--personality)
-            shift
-            PERSONALITY=$1
-            shift
-            ;;
-        -r|--product)
-            shift
-            PRODUCT=$1
-            shift
-            ;;
-        -i|--internal)
-            shift
-            BASE_URL="$INTERNAL_BASE_URL"
-            ;;
-        -h|--help)
-            usage
-            exit 0
-            ;;
-        --)
-            shift
-            break
-            ;;
-    esac
-done
+def check_call(*args):
+    log('$ ' + ' '.join(map(shlex.quote, args)))
+    subprocess.check_call(args, stdout=sys.stderr)
 
-log() {
-    echo >&2
-    echo "$@" >&2
-}
 
-download_and_verify() {
-  log "Downloading $1 from $2"
-  wget --continue $2 || return 1
+def ensure_keyring():
+    if os.path.isfile(KEYRING_SYSTEM_PATH):
+        return KEYRING_SYSTEM_PATH
 
-  log "Downloading signature for $1 from $3"
-  wget --continue $3 || return 1
+    keyring = os.path.abspath(KEYRING_FILENAME)
+    if not os.path.isfile(keyring):
+        log("Fetching Endless keyring")
+        check_call("wget", "-O", keyring, KEYRING_URL)
 
-  ensure_keyring
+    return keyring
 
-  log "Verifying..."
-  gpgv --keyring $KEYRING $(basename $3) $(basename $2) || return 1
 
-  log "Download ok."
-  return 0
-}
+def download_and_verify(image_url, signature_url):
+    log("Downloading image from", image_url)
+    check_call("wget", "--continue", image_url)
 
-ensure_keyring() {
-  if [ -f "$SYSTEM_KEYRING" ]
-  then
-    KEYRING="$SYSTEM_KEYRING"
-  else
-    KEYRING="$(readlink -f ./$KEYRING_FILENAME)"
-    if [ ! -f "$KEYRING" ]
-    then
-      log "Fetching Endless keyring"
-      wget -O "$KEYRING" "$KEYRING_URL"
-    fi
-  fi
-}
+    log("Downloading signature from", signature_url)
+    check_call("wget", "--continue", signature_url)
 
-log "Fetching manifest..."
-JSON=$(wget -O- $MANIFEST_BASE_URL/releases-$PRODUCT.json.gz | gzip -d)
+    keyring = ensure_keyring()
 
-# If the user specified a version, select that but otherwise we go with latest
-if test -z "$VERSION"
-then
-  SELECTED=$(echo $JSON | jq '.images[] | values' | jq -s 'max_by(.version)')
-else
-  SELECTED=$(echo $JSON | jq '.images[] | values' | jq -s ".[] | select(.version | contains(\"$1\"))")
-fi
+    log("Verifying")
+    check_call("gpgv", "--keyring", keyring,
+               os.path.basename(signature_url),
+               os.path.basename(image_url))
 
-if test -z "$PERSONALITY"
-then
-  PERSONALITY="base"
-fi
+    log("Download okay.")
+    return os.path.basename(image_url)
 
-# Only full images are supported for now
-OS_IMAGE_URL=$(echo $SELECTED | jq -r ".personality_images[\"$PERSONALITY\"] | .full | .file")
-OS_SIGNATURE_URL=$(echo $SELECTED | jq -r ".personality_images[\"$PERSONALITY\"] | .full | .signature")
 
-if test "null" = "$OS_IMAGE_URL" -o "null" = $OS_SIGNATURE_URL
-then
-  log "Problem fetching the image manifest. Please try again."
-  exit 1
-fi
+def main():
+    p = argparse.ArgumentParser(
+        description='Fetches and verifies an image (installer or regular OS)')
+    p.add_argument('-v', '--version',
+                   help='Image version (default: newest)')
+    p.add_argument('-p', '--personality',
+                   help='Image personality (default: base)',
+                   default='base')
+    p.add_argument('-r', '--product',
+                   help='Product (eg: eosinstaller; default: eos)',
+                   default='eos')
+    p.add_argument('-i', '--internal',
+                   action='store_true',
+                   help='Fetch images from the Endless internal network')
 
-# Download & verify
-download_and_verify "image" $BASE_URL/$OS_IMAGE_URL $BASE_URL/$OS_SIGNATURE_URL >&2
+    args = p.parse_args()
+    base = INTERNAL_BASE_URL if args.internal else PUBLIC_BASE_URL
 
-echo $(basename $OS_IMAGE_URL)
+    manifest_url = '{base}/releases-{product}.json.gz'.format(
+        base=base, product=args.product)
+    req = urllib.request.Request(manifest_url)
+    with urllib.request.urlopen(req) as conn:
+        with gzip.GzipFile(mode='r', fileobj=conn) as manifest_fp:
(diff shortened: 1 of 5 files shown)


eos-tech-support/eos-download-image
```diff
@@ -1,125 +1,114 @@
...
+
+
+def log(*args):
+    sys.stderr.write('\n')
+    print(*args, file=sys.stderr)
+    sys.stderr.flush()
+
+
+def check_call(*args):
+    log('$ ' + ' '.join(map(pipes.quote, args)))
```
dbnicholson: Nitpicky, but I think using `shlex.quote` would be nicer. `pipes` pulls in `shlex.quote`, but it's not in the documentation as far as I can tell - https://docs.python.org/3/library/pipes.html.

eos-tech-support/eos-download-image
```diff
@@ -1,125 +1,114 @@
...
+    try:
+        personality = personalities[args.personality]
+    except KeyError:
+        log("Personality", repr(args.personality), "not found")
+        log("Known personalities:", *sorted(personalities.keys()))
+        sys.exit(1)
+
+    image_url = '{}/{}'.format(base, personality['full']['file'])
+    signature_url = '{}/{}'.format(base, personality['full']['signature'])
+    print(download_and_verify(image_url, signature_url))
```
dbnicholson: Since all the calls in `download_and_verify` do their own logging, is the `print` necessary.

eos-tech-support/eos-download-image
```diff
@@ -1,125 +1,114 @@
...
+
+    manifest_url = '{base}/releases-{product}.json.gz'.format(
+        base=base, product=args.product)
+    req = urllib.request.Request(manifest_url)
+    with urllib.request.urlopen(req) as conn:
+        with gzip.GzipFile(mode='r', fileobj=conn) as manifest_fp:
+            manifest = json.load(codecs.getreader('utf8')(manifest_fp))
+
+    images = list(manifest['images'].values())
+    images.sort(key=lambda i: distutils.version.StrictVersion(i['version']))
```
dbnicholson: I wish we could use apt versioning here since that's what we use everywhere else, but it looks like `python3-apt` is not part of the core package list.

eos-tech-support/eos-download-image
wjt: The `print` is to stdout: this script is invoked by `eos-write-installer` and `eos-write-live`, which read the image filename from this script's stdout.

eos-tech-support/eos-write-live
```diff
@@ -0,0 +1,272 @@
...
+EOS_WRITE_IMAGE=$(dirname $0)/eos-write-image
+if [ ! -f $EOS_WRITE_IMAGE ]; then
+    EOS_WRITE_IMAGE='eos-write-image'
+fi
+EOS_DOWNLOAD_IMAGE=$(dirname $0)/eos-download-image
+if [ ! -f $EOS_DOWNLOAD_IMAGE ]; then
+    EOS_DOWNLOAD_IMAGE='eos-download-image'
+fi
+
+ARGS=$(getopt -o o:x:,lnmwfh -l "os-image:,windows-tool:,latest,ntfs,mbr,writable,force,help" -n "$0" -- "$@")
```
dbnicholson: The comma after `x:` in the short options looks incorrect to me. What does it do?

eos-tech-support/eos-write-live
```diff
@@ -0,0 +1,272 @@
...
+    exit 1
+}
+
+command -v unzip >/dev/null || {
+    echo "unzip is not installed... aborting!" >&2
+    echo "Try 'sudo apt-get install unzip'" >&2
+    exit 1
+}
+
+if test \! -b $DEVICE; then
```
dbnicholson: I think inside a script you can skip escaping `!` since `histexpand` is off in non-interactive shells. Doesn't hurt, though.

eos-tech-support/eos-write-live
```diff
@@ -0,0 +1,272 @@
...
+    PARTITION_SYSTEM_GUID="c12a7328-f81f-11d2-ba4b-00a0c93ec93b"
+    PARTITION_BIOS_BOOT_GUID="21686148-6449-6E6F-744E-656564454649"
+    PARTITION_BASIC_DATA_GUID="ebd0a0a2-b9e5-4433-87c0-68b6b72699c7"
+
+    # We want the data partition, "eosimages", to occupy all the space on the disk
+    # after the UEFI and BIOS boot partitions. But, we also want it to be numbered
+    # first: apparently Windows will only mount the partition numbered first,
+    # regardless of where it is on the disk.
+    DEVICE_EFI="${DEVICE}2"
+    DEVICE_BIOS="${DEVICE}3"
```
dbnicholson: This does make some assumptions about how the kernel driver labels partitions. I.e., with just the partition number and not with `p<part number` or any other scheme. Good enough for now, I think. There are probably other ways to do this, but here's one way I found:
```
partno=2
major=$(stat -c '%t' $DEVICE)
partdev="/dev/$(udevadm info -q name -p /sys/dev/block/${major}:${partno})"
```

eos-tech-support/eos-write-live
```diff
@@ -0,0 +1,272 @@
...
+# Give udisks a chance to notice the new partitions
+partprobe
+udevadm settle
+
+udisks_mount() {
+    udisksctl mount -b "$1" --no-user-interaction >&2
+    MOUNT_POINT=$(udisksctl info -b "$1" | grep "MountPoints:" | cut -f2 -d':')
+    test -d "$MOUNTPOINT"
+    echo $MOUNT_POINT
+}
```
dbnicholson: I guess you're using udisks for mounting to make sure that all mount points get unmounted if udisks decides to mount it. However, in that case, should you check if it's mounted first before calling `udiskctl mount`?

eos-tech-support/eos-download-image
dbnicholson: Right. I hope this doesn't break if we ever make a pre-release with `~` in the name. That doesn't seem to be allowed in any `distutils` versioning.

eos-tech-support/eos-download-image
dbnicholson: Got it.

eos-tech-support/eos-write-live
```diff
@@ -0,0 +1,272 @@
+#!/bin/bash -e
```
dbnicholson: This might be better named `eos-write-live-image`. `write-live` is a bit ambiguous to me.

dbnicholson:
Left a few comments. I don't understand all the details of `write-live`, but it seems to do what it says.

eos-tech-support/eos-download-image
wjt: Ah, `pipes.quote` is [apparently](https://docs.python.org/2.7/library/pipes.html#pipes.quote) a Python 2-ism. Thanks!

eos-tech-support/eos-write-live
wjt: Apparently it makes `-,` an acceptable option. I'll fix this and the other instance of it in `eos-write-installer`.

eos-tech-support/eos-write-live
wjt: It's neater without it; I'll change this and `eos-write-installer` (where this came from…)

eos-tech-support/eos-write-live
wjt: That's true, and it actually just bit me because I tried to write to a loop-mounted file. I'll look at fixing it.

eos-tech-support/eos-download-image
wjt: It does:

```python
>>> StrictVersion('1.0~pre1')
ValueError: invalid version number '1.0~pre1'
```

On the other hand:

```python
>>> u = LooseVersion('1.0~pre1')
>>> v = LooseVersion('1.0')
>>> u < v
False
```

which is unfortunate:

```bash
$ dpkg --compare-versions 1.0~pre1 lt 1.0; echo $?
0
```

I'm not sure what best to do about this. I'm sure that the [image list JavaScript](https://support.endlessm.com/hc/en-us/articles/209063006-Where-can-I-download-the-Endless-OS-image-files-directly-) will also get this wrong, but that's easier to fix after the fact.

wjt: I guess we could use LooseVersion so that at least the tool will not crash if we ever do that…

eos-tech-support/eos-write-live
wjt: Using `udisksctl` is inherited from `eos-write-installer`, to be honest. It's nice to not have to worry about inventing a mount point. I wouldn't expect the freshly-created partition to be mounted but I suppose it's possible…

eos-tech-support/eos-write-live
wjt: Renamed.

wjt:
I've addressed all bar two of your comments:

* checking whether the partition is already mounted: I don't expect it to be, it's freshly-created and there shouldn't be rules to auto-mount them
* version parsing: I'm not sure what best to do here. I suppose we could add a dependency on `python3-apt` but that would be a bit sad for the reasons discussed. The cop-out option would be to use `LooseVersion` and accept that the tool may pick the wrong image if we do make a release with a funky version.

wjt:
I also changed the image partition name to `eoslive`. This matches what the Windows tool creates, and helps disambiguate from old-style installer devices (which had an `eosimages` partition with different contents).

eos-tech-support/eos-write-image
```diff
@@ -52,14 +53,14 @@ while true; do
 done
 
 command -v dd >/dev/null || {
```
sgnn7: @wjt Please change these to `if`/`then`-style (`||` is really hard to read and maintain.

I think in other code we use something like `if [ "$(which dd)" == "" ]; then` for these command checking constructs.

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
+#!/bin/bash -e
+set -o pipefail
```
sgnn7: eos-write image uses a lot of pipes so this would fit there too

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+    MKFS_IMAGES="mkfs.ntfs"
+    MKFS_ARGS='--quick -L'
+    MKFS_PACKAGE=ntfs-3g
+else
+    MKFS_IMAGES="mkfs.exfat"
+    MKFS_ARGS=-n
+    MKFS_PACKAGE=exfat-utils
+fi
+
+command -v $MKFS_IMAGES >/dev/null || {
```
sgnn7: Ditto on command checking syntax. Also for this I'd make an array of required commands and loop through it instead of individual if/thens

eos-tech-support/eos-write-image
```diff
@@ -81,53 +82,59 @@ DD_IN_OPTS+=(bs="$BLOCK_SIZE")
...
+        dd if="$IMAGE" "${DD_IN_OPTS[@]}" | pv -s ${IMAGE_SIZE}
+    fi
+}
 
-if grep -qs $DEVICE /proc/mounts; then
-    # Protect against overwriting the device currently in use
-    echo "$DEVICE is currently in use -- please unmount and try again"
-    exit 1
-fi
+if [ $DEVICE == "-" ]; then
```
sgnn7: Quote your vars or you risk malicious escaping in passed-in vars

eos-tech-support/eos-write-image
```diff
@@ -81,53 +82,59 @@ DD_IN_OPTS+=(bs="$BLOCK_SIZE")
...
+    if [ ! -b $DEVICE ]; then
+        echo "$DEVICE not found or is not a block device" >&2
+        exit 1
+    fi
 
-if ! $FORCE; then
-    read -p "Are you sure you want to overwrite all data on $DEVICE? [y/N] "
-    response=${REPLY,,} # to lower
-    if [[ ! $response =~ ^(yes|y)$ ]]; then
+    if grep -qs $DEVICE /proc/mounts; then
```
sgnn7: Ditto on quoting vars

eos-tech-support/eos-write-image
```diff
@@ -81,53 +82,59 @@ DD_IN_OPTS+=(bs="$BLOCK_SIZE")
...
-    # Image is gzipped
-    # The following would calculate the original size
-    # of the unzipped file if less than 4G:
-    # IMAGE_SIZE=$(zcat -l "$IMAGE" | awk 'NR==2 { print $2 }')
-    # But this doesn't help for large images,
-    # so we just show relative progress unless .size is present
-    if [ -f "$IMAGE.size" ]; then
-        IMAGE_SIZE=$(cat $IMAGE.size)
-        PV_ARGS="-s $IMAGE_SIZE"
+    if ! $FORCE; then
```
sgnn7: Ditto on quoting vars

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+DEVICE="$1"
+
+USERID=$(id -u)
+if [ "$USERID" != "0" ]; then
+    echo "Program requires superuser privileges" >&2
+    exit 1
+fi
+
+# Check for required tools
+if $NTFS; then
```
sgnn7: Ditto on quoting vars - I'd use the `if[ ]; then` syntax here though.

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+    exit 1
+fi
+
+if grep -qs $DEVICE /proc/mounts; then
+    # Protect against overwriting the device currently in use
+    echo "$DEVICE is currently in use -- please unmount and try again" >&2
+    exit 1
+fi
+
+if ! $FORCE; then
```
sgnn7: Ditto on quoting vars - I'd use the `if[ ]; then` syntax here though.

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+    read -p "Are you sure you want to overwrite all data on $DEVICE? [y/N] "
+    response=${REPLY,,} # to lower
+    if [[ ! $response =~ ^(yes|y)$ ]]; then
+        exit 1
+    fi
+fi
+
+if [ ! -z "$FETCH_LATEST" ]; then
```
sgnn7: Ditto on quoting vars - Use `if[ ]; then`.

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+    if [ -z "$OS_IMAGE" ]; then
+        OS_IMAGE=$($EOS_DOWNLOAD_IMAGE --product eos)
+    fi
+    if [ -z "$WINDOWS_TOOL" ]; then
+        WINDOWS_TOOL=$($EOS_DOWNLOAD_IMAGE --windows-tool)
+    fi
+fi
+
+if [ ! -f "$WINDOWS_TOOL" -o ! -f "$OS_IMAGE" ]; then
```
sgnn7: Use `if [ condition_a ] || [ condition_b ]; then` instead of `-o` since it's more readable that way.

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+check_exists "$WINDOWS_TOOL" "Windows tool"
+
+echo
+echo "Partitioning $DEVICE..."
+
+find_by_type() {
+    echo "$1" | grep -i "type=$2" | cut -d' ' -f1
+}
+
+if $MBR; then
```
sgnn7: Use `if [ ]; then` here

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+    (
+        echo "1 : type=7"
+    ) | sfdisk --label dos $DEVICE
```
sgnn7: Why the strange `( )` syntax here?

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+
+    # We want the data partition, "eoslive", to occupy all the space on the disk
+    # after the UEFI and BIOS boot partitions. But, we also want it to be numbered
+    # first: apparently Windows will only mount the partition numbered first,
+    # regardless of where it is on the disk.
+    (
+        echo "2 : start=2048, size=62MiB, type=$PARTITION_SYSTEM_GUID"
+        echo "3 : size=1MiB, type=$PARTITION_BIOS_BOOT_GUID"
+        echo "1 : name=eoslive, type=$PARTITION_BASIC_DATA_GUID"
+    ) | sfdisk --label gpt $DEVICE
```
sgnn7: heredoc instead?

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+    partprobe $DEVICE
+    PARTMAP=$(sfdisk --dump $DEVICE)
+    DEVICE_IMAGES=$(find_by_type "$PARTMAP" "$PARTITION_BASIC_DATA_GUID")
+    DEVICE_EFI=$(find_by_type "$PARTMAP" "$PARTITION_SYSTEM_GUID")
+    DEVICE_BIOS=$(find_by_type "$PARTMAP" "$PARTITION_BIOS_BOOT_GUID")
+fi
+
+echo
+echo "Creating partitions"
+if ! $MBR; then
```
sgnn7: Ditto on brackets

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+$MKFS_IMAGES $MKFS_ARGS eoslive "${DEVICE_IMAGES}"
+
+# Give udisks a chance to notice the new partitions
+partprobe
+udevadm settle
+
+udisks_mount() {
+    udisksctl mount -b "$1" --no-user-interaction >&2
+    MOUNT_POINT=$(udisksctl info -b "$1" | grep "MountPoints:" | cut -f2 -d':')
+    test -d "$MOUNTPOINT"
```
sgnn7: if/then instead

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+    echo $MOUNT_POINT
+}
+
+udisks_unmount() {
+    udisksctl unmount -b "$1" --no-user-interaction >&2
+}
+
+if ! $MBR; then
```
sgnn7: Brackets ditto

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,284 @@
...
+        "$MOUNT_IMAGES_ENDLESS/"
+    echo "$(basename "${OS_IMAGE%.gz}")" > "${MOUNT_IMAGES_ENDLESS}/live"
+    cp "$WINDOWS_TOOL" "$MOUNT_IMAGES/"
+fi
+
+"$EOS_WRITE_IMAGE" "$OS_IMAGE" "-" > "${MOUNT_IMAGES_ENDLESS}/endless.img"
+udisks_unmount "$DEVICE_IMAGES"
+
+if $MBR; then
+    unzip -p "${BOOT_ZIP}" "ntfs/boot.img" | dd of="${DEVICE}" bs=446 count=1
```
sgnn7: Some comments here about the offsets would be nice as well as ones for settle/sync

eos-tech-support/eos-download-image
dbnicholson: It hasn't happened, so maybe we just don't care.

Another option is to copy `NativeVersion` from https://anonscm.debian.org/git/pkg-python-debian/python-debian.git/tree/lib/debian/debian_support.py#n218. Maybe even:
```
try:
    from debian.debian_support import Version
except ImportError:
    # roll our own Version class using NativeVersion
```
I would definitely support including python3-apt and python3-debian in the core as they're small and I use them often. But I don't know how much effort to really put in here.

eos-tech-support/eos-write-live
dbnicholson: I might be wrong, but I actually do think that if the device is a USB stick, udisks will try to automount it from a graphical session once valid filesystems show up. I've definitely seen QA struggle with that and use stuff like `sudo umount -a` to try to clean up with hilarious side effects.

I think it's probably fine as is, but one way to handle this would be to preemptively unmount:
```
udisksctl unmount -b "$1" --no-user-interaction 2>/dev/null || true
```

eos-tech-support/eos-write-image
dbnicholson: Since `$FORCE` is controlled within this script (it's either `true` or `false`), then I think this usage is entirely appropriate and more direct than testing the string.

sgnn7: Fair enough - didn't see that this was set internally.

eos-tech-support/eos-write-live-image
dbnicholson: As mentioned in the other comment, I `NTFS` is controlled and using it as a boolean is entirely appropriate to me.

wjt:
@sgnn7 I think I addressed all your points!

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,292 @@
...
+if [ $# -lt 1 ] ; then
+    echo "Missing DEVICE" >&2
+    usage >&2
+    exit 1
+fi
+
+DEVICE="$1"
+
+USERID=$(id -u)
+if [ "$USERID" != "0" ]; then
```
sgnn7: If we assume it's bash, I think you can use `$EUID`

sgnn7:
@wjt Looks good to me but having @dbnicholson have a final lookover would be good so I'll leave it unmerged.

eos-tech-support/eos-write-live-image
```diff
@@ -0,0 +1,291 @@
...
+    # image.
+    sfdisk --label gpt "$DEVICE" <<EFI_PARTITIONS
+2 : start=2048, size=62MiB, type=$PARTITION_SYSTEM_GUID
+3 : size=1MiB, type=$PARTITION_BIOS_BOOT_GUID
+1 : name=eoslive, type=$PARTITION_BASIC_DATA_GUID
+EFI_PARTITIONS
+    udevadm settle
+    partprobe "$DEVICE"
+    PARTMAP=$(sfdisk --dump "$DEVICE")
+    DEVICE_IMAGES=$(find_by_type "$PARTMAP" "$PARTITION_BASIC_DATA_GUID")
```
wjt: @dbnicholson Here's the approach I took to finding the device nodes for partitions.

In another script I've used `lsblk` to go in the other direction (partition device → whole-disk device) but I think looking them up by type in `sfdisk`'s output is easier here, and reasonable given that we just created them that way.

dbnicholson: Yeah, seems good. I wish there was a more straightforward way to do this. Here's another way I found with `lsblk`:
```
$ DEVICE=/dev/sda
$ GUID=c12a7328-f81f-11d2-ba4b-00a0c93ec93b
$ part=$(lsblk -o PARTTYPE,NAME -nl "$DEVICE" | awk "{if (\$1 == \"$GUID\") print \$2}")
$ echo $part
sda1
```

dbnicholson:
Looks good, thanks for working through the comments!

wjt:
Thanks for all the feedback, both!

PR #299, opened 2016-08-29
base-depends: add sound-theme-freedesktop

cosimoc:
This was lost at some point during the package reorganization.

https://phabricator.endlessm.com/T12866

### DIFF
--- a/base-depends
+++ b/base-depends
@@ -29,5 +29,6 @@ locales
 lzma
 mawk
 netbase
+sound-theme-freedesktop
 ttf-freefont
 xdg-utils


dbnicholson:
It looks like this used to be a dependency of canberra in 2.6:
```
$ apt-cache rdepends sound-theme-freedesktop
sound-theme-freedesktop
Reverse Depends:
  Suggests: kde-runtime
  Depends: gnome-core
  Depends: freedesktop-sound-theme
$ chdist apt-cache eos2.6 rdepends sound-theme-freedesktop
sound-theme-freedesktop
Reverse Depends:
  Depends: libcanberra0
  Depends: gnome-core
  Depends: freedesktop-sound-theme
  Recommends: empathy
```
So we were getting it implicitly from libcanberra0 before.

PR #300, opened 2016-09-08
Pre-seed readahead file on amd64 architectures

jprvita:
This makes the first boot ~20s faster, on average.

https://phabricator.endlessm.com/T5564


cosimoc:
Contents of `/var` are not deployed through ostree, so this won't have any impact on our images.
Ideally I think we could ship the pre-seeded readahead file with the OSTree itself in `/usr`, since all the things that are read at startup time are really part of OSTree.

jprvita:
Oh right, the `/var` contents are not the ones that come through OSTree.

I have thought about shipping it with the OSTree in its default localtion `/.readahead`, but the problem is that this file is also updated on every boot with the current boot file access pattern, so we need it in a writable location (same applies for `/usr`). We can disable the `systemd-readahead-collect.service` which updates the file, but I believe in this case we need to somehow trim the model-specific portions of it (kernel modules, for example).

PR #306, opened 2016-09-19
eos-write-*: Remove fallback.efi from removable devices

carlocaione:
https://phabricator.endlessm.com/T13374

### DIFF
--- a/eos-tech-support/eos-write-image
+++ b/eos-tech-support/eos-write-image
@@ -5,7 +5,7 @@ set -o pipefail
 # to a device (e.g., USB drive or SD card)
 # using dd, with status monitoring via pv
 
-ARGS=$(getopt -o fh -l "force,sparse,debug,help" -n "$0" -- "$@")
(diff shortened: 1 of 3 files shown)


eos-tech-support/eos-write-image
```diff
@@ -140,4 +146,42 @@ else
     # Try to format / discard all the sectors on the device
     blkdiscard "$DEVICE" &>/dev/null || true
     cat_image | dd of="$DEVICE" "${DD_OUT_OPTS[@]}"
+
+    # If removable device, delete $ESP/EFI/BOOT/fallback.efi
+    if [ "$REMOVABLE" ]; then
```
ramcq: Can we learn somehow from /sys if a device is removable? People also use eos-write-image with /dev/sdX to make USB sticks etc.

dsd: The short answer here is "no". I suggest sticking with the command line arg.

PR #307, opened 2016-09-19
eos-diagnostics: add sys_vendor value

dsd:
For code that uses DMI info to match specific systems, the most common
fields used are sys_vendor and product_name together.

sys_vendor was not being reported here, add it.

@cosimoc can you take a quick look?


cosimoc:
@dsd Looks good, merged.

PR #308, opened 2016-09-21
eos-diagnostics: include codecs info

rshuler:
By listing the contents of /var/lib/codecs, we can easily see
which keys are installed and which codecs have been decrypted.

https://phabricator.endlessm.com/T13401

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -293,6 +293,16 @@ function dumpDiagnostics(filename) {
     fullDump += collectChromiumPluginsInfo();
     fullDump += '\n';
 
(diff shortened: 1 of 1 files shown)


mariospr:
I agree that the output of find is ok to know which codecs are installed and decrypted, but it would still be worth knowing the output of `gst-inspect-1.0`, so that we are also certain about what exactly GStreamer has recognized and loaded properly.

Actually, it would also be nice to include the output of `gst-inspect-1.0 -b` too, since sometimes there are issues with plugins that get blacklisted for obscure reasons, which is sometimes not obvious to know right from the symptoms of a bug report.

PR #312, opened 2016-09-23
eos-write-image: Move grubx64.efi on removable devices

carlocaione:
https://phabricator.endlessm.com/T13374


dsd:
Can you note in the commit message why the previous approach didn't work?

carlocaione:
@dsd done.

PR #313, opened 2016-09-28
os-depends: Prefer linux-signed-image on amd64

dbnicholson:
For amd64, prefer the linux-signed-image variant, which contains the
kernel signed for UEFI.

https://phabricator.endlessm.com/T12944


dsd:
I guess I need to hold off accepting this until a signed kernel package is present in the repos?

dbnicholson:
Ah, you're right. I was thinking it would skip `linux-signed-image-generic`, but that exists in the repo. It's dependency doesn't yet.

PR #314, opened 2016-09-30
eos-download-image: fix fetching boot.zip[.asc]

wjt:
I forgot about this script when I changed the JSON format!

https://phabricator.endlessm.com/T13224


wjt:
The change in question is endlessm/eos-build@aa442fa, part of endlessm/eos-build#409.

PR #318, opened 2016-10-04
T11518 add .xz support to eos-write-* tools

ramcq:
https://phabricator.endlessm.com/T11518

### DIFF
--- a/eos-tech-support/eos-write-image
+++ b/eos-tech-support/eos-write-image
@@ -62,7 +62,6 @@ done
 declare -A dependencies
 dependencies=(
     [dd]='coreutils'
-    [pv]='pv'
     [xzcat]='xz-utils'
     [zcat]='gzip'
 )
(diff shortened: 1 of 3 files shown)


eos-tech-support/eos-write-image
```diff
@@ -105,8 +105,22 @@ if [ ! -f "$IMAGE" ]; then
...
+    echo "Writing..." >&2
+    cat
+    echo "Done!" >&2
+  }
+fi
+
 cat_image() {
-    if file --mime-type "$IMAGE" | grep -q gzip$; then
+    if file --mime-type "$IMAGE" | grep -q xz$; then
+        # Image is xzipped, xz has it's own rate meter
```
wjt: Should be *its*.

eos-tech-support/eos-write-installer
```diff
@@ -97,7 +97,7 @@ dependencies=(
     [dd]='coreutils'
     [gzip]='gzip'
     [mkfs.exfat]='exfat-utils'
-    [pv]='pv'
+    [xz]='xz-utils'
```
wjt: Shouldn't this be `xzcat` again?

ramcq: It was already checking for gzip rather than zcat so I thought I'd be consistent. Have changed them all so we're instead correct, rather than merely consistent.

eos-tech-support/eos-write-image
ramcq: Fixed.

PR #321, opened 2016-10-05
eos-convert-system: Operate on the correct /var/lib/flatpak

dbnicholson:
The /var/lib/flatpak symlink is changed after everything has been copied
to /sysroot. So, changing the running /var/lib/flatpak has no effect
after rebooting. Check and change the correct symlink in
/sysroot/var/lib/flatpak.

https://phabricator.endlessm.com/T13748


dbnicholson:
Wrong issue.

PR #323, opened 2016-10-07
Add uuid-dev to the SDK dependencies

ebassi:
https://phabricator.endlessm.com/T13157


ptomato:
**CR pass**, we keep going...

PR #330, opened 2016-10-17
eos-write-live-image: Use grub-bios-setup to write bootloader images

carlocaione:
https://phabricator.endlessm.com/T13590

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -147,6 +147,7 @@ dependencies=(
     [unzip]='unzip'
     [xzcat]='xz-utils'
     [zcat]='gzip'
+    [grub-bios-setup]='grub-pc'
 )
 
 if [ "$ISO" ]; then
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-write-live-image
```diff
@@ -361,11 +362,10 @@ else
...
         unzip -q -p "${BOOT_ZIP}" "ntfs/core.img" | dd of="${OUTPUT}" bs=512 seek=1
     else
-        # Bootstrap code, built to jump to the offset of BIOS boot partition
-        unzip -q -p "${BOOT_ZIP}" "live/boot.img" | dd of="${OUTPUT}" bs=446 count=1
-        udevadm settle  # udev recreates device files after any write to the device
-        # The rest of GRUB goes into a dedicated BIOS-boot partition
-        unzip -q -p "${BOOT_ZIP}" "live/core.img" | dd of="${DEVICE_BIOS}" bs=512
+        DIR_LIVE="$(mktemp -d -t grub_dir_live.XXXXXX)"
+        unzip -q -d "${DIR_LIVE}" "${BOOT_ZIP}" "live/*"
+        /usr/sbin/grub-bios-setup --directory="${DIR_LIVE}" "${OUTPUT}"
```
wjt: We could use `grub-bios-setup` on the `$MBR` path above, too, if we were feeling keen… it would actually be more consistent with what the Windows tool does (since it now basically invokes an internal copy of `grub-bios-setup`).

PR #332, opened 2016-10-18
eos-list-users / eos-reset-password: Update for eos3

rshuler:
The hard-coded /dev/sda2 is no longer valid for eos3.

Instead, let's list all users on all ostree devices,
and then require the user to explicitly specify the device
that contains the user whose password is to be reset.

With these changes, we now support split disks, too.

https://phabricator.endlessm.com/T6324

### DIFF
--- a/eos-tech-support/eos-list-users
+++ b/eos-tech-support/eos-list-users
@@ -1,9 +1,6 @@
 #!/bin/bash -e
 
-# Lists all users (i.e., all home directories on /dev/sda2)
-
-DEVICE=/dev/sda2
-MOUNT=/mnt/sda2
+# Lists all users (i.e., all home directories on all ostree devices)
 
 USERID=$(id -u)
 if [ "$USERID" != "0" ]; then
@@ -11,17 +8,34 @@ if [ "$USERID" != "0" ]; then
     exit 1
 fi
 
+MOUNT=`mktemp -d`
+
 cleanup() {
     set +e
-    umount $MOUNT
+    if mount | grep $MOUNT; then
+        umount $MOUNT
+    fi
     rmdir $MOUNT
 }
 trap cleanup EXIT
 
 mkdir -p $MOUNT
-mount $DEVICE $MOUNT
 
-# FIXME A more robust approach would be the following:
-# - read /usr/etc/login.defs and extract UID_MIN
-# - read /etc/passwd and get all users with uid >= UID_MIN
-ls $MOUNT/home
+DEVICES=`blkid -t LABEL=ostree | awk -F ":" '{print $1}'`
+
+echo
+
+for DEVICE in $DEVICES
+do
+    mount $DEVICE $MOUNT
+
+    # FIXME A more robust approach would be the following:
+    # - read /usr/etc/login.defs and extract UID_MIN
+    # - read /etc/passwd and get all users with uid >= UID_MIN
+    echo Users on $DEVICE:
+    ls $MOUNT/home
+    echo
+
+    umount $MOUNT
+done
+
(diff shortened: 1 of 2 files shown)


eos-tech-support/eos-list-users
```diff
@@ -1,27 +1,41 @@
...
     echo "Program requires superuser privileges"
     exit 1
 fi
 
+MOUNT=`mktemp -d`
+
 cleanup() {
     set +e
-    umount $MOUNT
+    if mount | grep $MOUNT; then
```
cosimoc: Perhaps just `umount $MOUNT || true` here would do the trick?

eos-tech-support/eos-list-users
```diff
@@ -1,27 +1,41 @@
...
 trap cleanup EXIT
 
 mkdir -p $MOUNT
-mount $DEVICE $MOUNT
 
-# FIXME A more robust approach would be the following:
-# - read /usr/etc/login.defs and extract UID_MIN
-# - read /etc/passwd and get all users with uid >= UID_MIN
-ls $MOUNT/home
+DEVICES=`blkid -t LABEL=ostree | awk -F ":" '{print $1}'`
```
cosimoc: When running this on my dev-converted laptop, it looks like it printed two devices with the label:
```
/dev/block/8
/dev/sda3
```

Did you verify it does not happen on a regular system? I guess it's OK to list them both, but it may be a bit confusing if they actually are the same.

eos-tech-support/eos-reset-password
```diff
@@ -32,16 +30,16 @@ else
...
+
 cleanup() {
     set +e
-    umount $MOUNT/boot
     umount $MOUNT
     rmdir $MOUNT
 }
 trap cleanup EXIT
 
 mkdir -p $MOUNT
```
rshuler: Right.  I'll remove that.

eos-tech-support/eos-list-users
rshuler: I tried that first, but it still spits out an error message (and because of the loop, the normal case is that the device has been unmounted upon a clean exit)

eos-tech-support/eos-reset-password
```diff
@@ -4,15 +4,13 @@
 # Except for the special shared account,
 # the user will be asked for a password on next login
 
-BOOTDEV=/dev/sda1
-DEVICE=/dev/sda2
-MOUNT=/mnt/sda2
-
-if [ -z "$1" ]; then
-    echo usage: $0 username
+if [ -z "$2" ]; then
```
rshuler: Makes sense.

eos-tech-support/eos-list-users
rshuler: I tested by running from USB on an EC-100 eMMC running an ostree image, and did not see any /dev/block/* devices.  I also just tested locally on my dev-converted laptop and I still don't see any /dev/block/* device:

```
roddy@endless:~/checkout/eos-meta/eos-tech-support$ sudo ./eos-list-users 
[sudo] password for roddy: 

Users on /dev/sda3:
roddy  shared

roddy@endless:~/checkout/eos-meta/eos-tech-support$ 
```

eos-tech-support/eos-list-users
cosimoc: I think this should work
```
umount $MOUNT > /dev/null 2>&1 || true
```

eos-tech-support/eos-list-users
cosimoc: Odd, I can't seem to reproduce it here either now.

eos-tech-support/eos-list-users
rshuler: Yes, I thought of that, but if the umount fails for some other reason, we would not see the error message.  On the other hand, perhaps there is no practical other reason why umount would fail that we would care about, but still I think it is better to not hide potentially real error messages.

PR #334, opened 2016-10-21
eos-diagnostics: Check the right VERSION files for chromium plugins

mariospr:
https://phabricator.endlessm.com/T13464

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -127,23 +127,41 @@ function collectTemperatureInfo() {
 }
 
 function collectChromiumPluginsInfo() {
-    let output = 'Version: ';
-
-    let flashPath = Gio.file_new_for_path('/var/lib/chromium-plugins-extra/eos-chrome-plugin-updater/libpepflashplayer.so');
-    if (!flashPath.query_exists(null)) {
-	output += 'not installed\n';
-	return output;
-    }
-
+    let oldVersion = null;
     try {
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-diagnostics
```diff
@@ -127,23 +127,35 @@ function collectTemperatureInfo() {
...
-    let output = 'Version: ';
-
+    let flashOutput = 'Flash Version: ';
     let flashPath = Gio.file_new_for_path('/var/lib/chromium-plugins-extra/eos-chrome-plugin-updater/libpepflashplayer.so');
-    if (!flashPath.query_exists(null)) {
-	output += 'not installed\n';
-	return output;
+    if (flashPath.query_exists(null)) {
+        try {
+	    let [res, version] = GLib.file_get_contents('/var/lib/chromium-plugins-extra/eos-chrome-plugin-updater/Flash_VERSION.txt');
```
wjt: I guess everyone with this new version of `eos-diagnostics` will have run the new updater, so there's no need to fall back to the old `VERSION.txt` file?

mariospr: Yes, that was my thinking, although that won't apply to people who had the plugins applied with the previous version of the script and became offline after the OS upgrade.

I think it's not a very likely scenario, but maybe it's worth keeping the old code as a fallback.

mariospr: I updated the PR, please review. Thanks!

PR #339, opened 2016-11-03
Symlinks allowing Flatpak app bundles to provide a 32-bit userland

aperezdc:
https://phabricator.endlessm.com/T13322
https://phabricator.endlessm.com/T13865

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -4,8 +4,27 @@
 # Uncomment this to turn on verbose mode.
 #export DH_VERBOSE=1
 
+DEB_TARGET_ARCH := $(shell dpkg-architecture -q DEB_TARGET_ARCH)
+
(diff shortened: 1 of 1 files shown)


aperezdc:
[This long comment](https://phabricator.endlessm.com/T13865#284538) is a good read to understand the rationale for this patch. The TL;DR is that this allows an application bundle to provide a 32-bit userland _now_, and when Flatpak implements the moving parts needed to support this, the changes from this PR can be reverted _without needing to rebuild the application bundles_.

debian/changelog
```diff
@@ -1,3 +1,15 @@
+eos-meta (1.5.1) UNRELEASED; urgency=medium
```
cosimoc: We do not touch the changelog -- please put these notes in the commit message instead.

PR #340, opened 2016-11-03
Add s905x metapackages

carlocaione:
https://phabricator.endlessm.com/T14004


dbnicholson:
You need to add the new metapackages to debian/control for this to be used.

PR #346, opened 2016-11-18
eos-diagnostics: dump CPU info

dsd:
Add CPU information to the regular and hardware dumps.

@cosimoc can you review this please? should be ultra quick


cosimoc:
@dsd Looks good!

PR #348, opened 2016-11-28
eos-sdk-runtime-depends: add knowledge content dev package

mattdangerw:
https://phabricator.endlessm.com/T13418


xjuan:
CR pass

PR #349, opened 2016-11-29
eos-{list-users,reset-password}: support dual-boot installations

wjt:
I'm not particularly happy with the code duplication here. One option would be
to merge the two scripts (and preferably rewrite it in Python in the process --
I had some nasty scoping problems with the cleanup hooks).

https://phabricator.endlessm.com/T14302

### DIFF
--- a/eos-tech-support/eos-reset-password
+++ b/eos-tech-support/eos-reset-password
@@ -7,7 +7,7 @@
 if [ "$#" -ne 2 ]; then
     echo "usage: $0 device username"
     echo "(run eos-list-users to see a list of devices and usernames)"
-    exit
+    exit 1
 fi
 DEVICE=$1
 USERNAME=$2
@@ -20,7 +20,7 @@ fi
 
 if [ $USERNAME = root ]; then
     echo "Resetting root password not allowed"
-    exit
+    exit 1
 elif [ $USERNAME = shared ]; then
     # For the shared account, set the change date (days since Jan 1, 1970)
     # to today (so that user will not be asked to set a password)
--- a/eos-tech-support/eos-reset-password
+++ b/eos-tech-support/eos-reset-password
@@ -31,14 +31,49 @@ else
 fi
 
 MOUNT=`mktemp -d`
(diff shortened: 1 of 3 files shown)


eos-tech-support/eos-reset-password
```diff
@@ -31,14 +31,49 @@ else
...
+    rmdir "$MOUNT"
 }
 trap cleanup EXIT
 
 mount $DEVICE $MOUNT
+
+# Check if it's a Windows partition with a dual-boot Endless OS image
+LABEL=$(lsblk -o LABEL --raw --nodeps --noheadings "$DEVICE")
+ENDLESS_IMG="$MOUNT/endless/endless.img"
+if [ "$LABEL" != "ostree" ] && [ -e "$ENDLESS_IMG" ]; then
```
mariospr: I think you should check whether `$LABEL` is an empty string to protect against bogus runs of `lsblk`

wjt: `$LABEL` can legitimately be the empty string. Here's the `lsblk` output from a modern.ie Windows 7 VM:

screenshot from 2016-11-30 11-25-21

Notice that `sda2`, the partition to which Endless OS is installed, has an empty label.

This script could explicitly handle the case where `$DEVICE` is neither an `ostree` partition, nor contains `/endless/endless.img`, but I assume that the call to `ostree` will give a meaningful error message in that case. What do you think?

mariospr: > `$LABEL` can legitimately be the empty string. Here's the `lsblk` output from a modern.ie Windows 7 VM:
> Notice that `sda2`, the partition to which Endless OS is installed, has an empty label.

Sounds good then, thanks for the clarification.

> This script could explicitly handle the case where `$DEVICE` is neither an `ostree` partition, nor contains `/endless/endless.img`, but I assume that the call to `ostree` will give a meaningful error message in that case. What do you think?

I agree it's probably not needed to handle that case, I re-read the code below and I think it should work fine even without handling that case explicitly.

PR #357, opened 2017-01-05
Get repo urls from ostree call, rather than from awk parsing

rmacqueen:
The awk command was incorrect because there are an arbitrary number
of words in between the repo name and the url, so trying to extract
the url from that line was giving us incorrect results. Instead
let's just hardcode the repo names since they aren't going to change
anytime soon.

### DIFF
--- a/eos-tech-support/eos-convert-system-qa
+++ b/eos-tech-support/eos-convert-system-qa
@@ -93,14 +93,20 @@ if $OSTREE || $APPS; then
 
     # Change the Flatpak runtime and apps server URLs.
     if $APPS; then
-        runtime_url=$(flatpak remote-list -d | awk '/^eos-runtimes/{print $3}')
+        if [ -d /var/endless-extra/flatpak/repo ];
+            flatpak_repo=/var/endless-extra/flatpak/repo
+        else
+            flatpak_repo=/var/lib/flatpak/repo
+        fi
+
+        runtime_url=$(ostree --repo="$flatpak_repo" config get 'remote "eos-runtimes".url')
         runtime_repo=${runtime_url##*/}
         new_runtime_url="https://${OSTREE_USER}:${OSTREE_PASSWORD}@${OSTREE_HOST}/${OSTREE_DEV_ROOT}/${runtime_repo}"
 
         echo "Setting flatpak runtimes URL to $new_runtime_url"
         flatpak remote-modify eos-runtimes --url="$new_runtime_url"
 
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-convert-system-qa
```diff
@@ -93,16 +93,14 @@ if $OSTREE || $APPS; then
 
     # Change the Flatpak runtime and apps server URLs.
     if $APPS; then
-        runtime_url=$(flatpak remote-list -d | awk '/^eos-runtimes/{print $3}')
-        runtime_repo=${runtime_url##*/}
-        new_runtime_url="https://${OSTREE_USER}:${OSTREE_PASSWORD}@${OSTREE_HOST}/${OSTREE_DEV_ROOT}/${runtime_repo}"
+        RUNTIME_REPO=eos-amd64
```
dbnicholson: This won't work on arm.

dbnicholson: I think we should just improve the parsing. We can also use ostree to get it directly: `ostree config get 'remote "eos-runtimes".url'`.

eos-tech-support/eos-convert-system-qa
```diff
@@ -93,14 +93,14 @@ if $OSTREE || $APPS; then
 
     # Change the Flatpak runtime and apps server URLs.
     if $APPS; then
-        runtime_url=$(flatpak remote-list -d | awk '/^eos-runtimes/{print $3}')
+        runtime_url=$(ostree config get 'remote "eos-runtimes".url')
```
dbnicholson: Great. There is one little issue, though. If this is a split disk, a different repo is used. Can you pass the repo explicitly? Something like this:
```
if [ -d /var/endless-extra/flatpak/repo ]; then
    flatpak_repo=/var/endless-extra/flatpak/repo
else
    flatpak_repo=/var/lib/flatpak/repo
fi
runtime_url=$(ostree --repo="$flatpak_repo" config get 'remote "eos-runtimes".url')
...
```
And continue using `--repo="$flatpak_repo"` while operating on the apps.

rshuler:
@rmacqueen Are you going to be able to finish this up?

@abnerf Please note that in the meantime, eos-convert-system-qa is broken in that it sets up the flatpak repos incorrectly.

rmacqueen:
Yea I can. Got swamped with app stuff the last couple days but will try to get to it tomorrow

rmacqueen:
Ready again

PR #358, opened 2017-01-05
core: add telepathy packages

GeorgesStavracas:
The Telepathy packages were removed in the past because
Empathy itself wasn't very used. With the upcoming GNOME
repositories enabled by default, however, Polari will
require Telepathy - and will fail without absolutely
no error message if Telepathy isn't there.

To make Polari work properly and avoid further frustrations,
add Telepathy (but not Empathy) back.

https://phabricator.endlessm.com/T14703

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -117,6 +117,9 @@ smbclient
 strace
 system-config-printer-gnome
 systemd-coredump
+telepathy-idle
+telepathy-logger
+telepathy-mission-control-5
 totem
 tracker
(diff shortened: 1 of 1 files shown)


rshuler:
Do we really need to list all of these packages?  Is there maybe just one package that is actually required, and we can let it install its dependencies?  I recall someone saying that the mission control was the required package (which would pull in a couple of the others I think).  Let's not install others unless they are truly required.

PR #363, opened 2017-01-12
ISO: Ship a squashfs image

carlocaione:
https://phabricator.endlessm.com/T13560

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -153,6 +153,7 @@ if [ "$ISO" ]; then
         [mcopy]='mtools'
         [mkisofs]='genisoimage'
         [grub-mkimage]='grub-common'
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-write-live-image
```diff
@@ -153,6 +153,7 @@ if [ "$ISO" ]; then
         [mcopy]='mtools'
         [mkisofs]='genisoimage'
         [grub-mkimage]='grub-common'
+        [mksquashfs]='squashfs-tools'
```
wjt: In theory we should add xorriso, dosfstools, mtools, genisoimage, squashfs-tools to the package dependencies, or else this path of the script will not work on unconverted Endless OS. But I'd expect ISO generation to move into the image builder in due course, so this would be an unnecessary 3MB in the image.

PR #366, opened 2017-01-13
add eos-transmission-seeder script

ramcq:
Derivative of eos-download-image which:
 - grabs the releases JSON for a given product
 - finds the torrents for the selected version & personalities
 - uses transmissionrpc to add them to a Transmission daemon
 - stops (and optionally deletes) any other torrents of the same product

Intended to be set up on cron (or similar) to set up a local mirror of images you care
about, help Endless seed their torrents, or both.

https://phabricator.endlessm.com/T15043

### DIFF
--- a/eos-tech-support/eos-transmission-seeder
+++ b/eos-tech-support/eos-transmission-seeder
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+# vim: fileencoding=utf-8 sts=4 sw=4 et
+import argparse
+import codecs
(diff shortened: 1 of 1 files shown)


wjt:
I know this is a really old branch but: I wonder if this might usefully just be a different mode of `eos-download-image` rather than copy-paste? Particularly if we might add a partial version of `eos-download-image --mirror` which mirrors the last 2 releases and deletes old one – there'd be even more overlap.

PR #367, opened 2017-01-13
Xaiki/t15040 ndn

xaiki:
This integrates NFD and EDD (python-eos-data-distribution) that will enable our NDN updater to work.


dbnicholson:
Should these be part of the runtime, too, or just the OS?

magcius:
Just the OS.

xaiki:
just the OS i guess, i don't think any of our flatpaks can make use of it now, we'll move it later if needed.

PR #368, opened 2017-01-17
eos-platform-runtime-depends: added mali drivers dep

xjuan:
Added mali dependency on flatpak runtime

https://phabricator.endlessm.com/T15102


ptomato:
According to Rob, "That fix is the correct implementation of the improper approach which was agreed for 3.0", so I think we should merge this.

**CR pass**

To be merged Friday.

PR #371, opened 2017-01-18
eos-download-image: handle python <3.3 by making shlex.quote use optional

ramcq:
Fix running the eos-download-image script on older distros.

https://phabricator.endlessm.com/T15135

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -25,7 +25,11 @@ def log(*args):
 
 
 def check_call(*args):
-    log('$ ' + ' '.join(map(shlex.quote, args)))
+    if hasattr(shlex, 'quote'):
+        qargs = map(shlex.quote, args)
+    else:
+        qargs = args
+    log('$ ' + ' '.join(qargs))
     subprocess.check_call(args, stdout=sys.stderr)
 
 


eos-tech-support/eos-download-image
```diff
@@ -25,7 +25,11 @@ def log(*args):
 
 
 def check_call(*args):
-    log('$ ' + ' '.join(map(shlex.quote, args)))
+    if shlex.quote:
+        qargs = map(shlex.quote, args)
+    else:
+        qargs = args
```
dbnicholson: I think you should make this `map(str, args)` in case any non string args are passed in. Otherwise you'll bomb below in the `join`. I have hit this before trying to make such a `check_call` wrapper.

eos-tech-support/eos-download-image
```diff
@@ -25,7 +25,11 @@ def log(*args):
 
 
 def check_call(*args):
-    log('$ ' + ' '.join(map(shlex.quote, args)))
+    if shlex.quote:
```
dbnicholson: Wait, does this work? Won't this just throw `AttributeError` if it doesn't exist? I think you want `hasattr(shlex, 'quote')`.

eos-tech-support/eos-download-image
dbnicholson: Actually, it appears that `shlex.quote` bombs if the args aren't strings, so maybe it doesn't matter:
```
>>> shlex.quote(['foo', 1, 2])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.4/shlex.py", line 285, in quote
    if _find_unsafe(s) is None:
TypeError: expected string or buffer
```

eos-tech-support/eos-download-image
ramcq: Doh... fixed this. Also I now have a Linux Educacional VM so I can test as well. Cunning!

PR #373, opened 2017-01-20
T15177 coredumps

pwithnall:
Two commits to add `systemd-coredump` support to `eos-diagnostics`.

https://phabricator.endlessm.com/T15177

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -570,6 +570,11 @@ let diagnostics = [
         hardwareInfo: true,
         content: function() { return trySpawn('journalctl --no-pager'); },
     },
+    {
+        title: 'Coredump',
+        hardwareInfo: true,
+        content: function() { return trySpawn('coredumpctl list'); },
+    },
 ];
 
 function dumpDiagnostics(filename) {
@@ -624,6 +629,28 @@ function copyAcpiFirmware(outputPath)
     }
 }
 
+function copyCoredumps(outputPath)
+{
+    let dir = Gio.File.new_for_path('/var/lib/systemd/coredump');
+    let output = Gio.File.new_for_path(outputPath);
+    let fileEnum;
+
+    try {
+        fileEnum = dir.enumerate_children('standard::name',
+                                          Gio.FileQueryInfoFlags.NONE, null);
+    } catch (e) {
+        return;
+    }
+
+    let info;
+    while ((info = fileEnum.next_file(null))) {
+        let name = info.get_name();
+
+        let target = output.resolve_relative_path('coredump_' + name.toLowerCase());
+        fileEnum.get_child(info).copy(target, 0, null, null);
+    }
+}
+
 function dumpToArchive(filename, contentsFunc) {
     let dir = '/tmp/' + filename;
     GLib.mkdir_with_parents(dir, parseInt('0700', 8));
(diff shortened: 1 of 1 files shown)


pwithnall:
Branch updated to split coredump support out into a separate `--coredumps` argument.

eos-tech-support/eos-diagnostics
```diff
@@ -556,7 +556,7 @@ let diagnostics = [
     {
         title: 'Journal',
         hardwareInfo: true,
-        content: function() { return trySpawn('journalctl --no-pager'); },
+        content: function() { return trySpawn('journalctl --no-pager --output verbose'); },
```
rshuler: I don't think this is such a good idea.  I just checked on a system that I rebooted a few minutes ago, and this would bloat the diagnostic file size from 2688 lines / 277 kB to 46241 lines / 1.9 MB.  And, looking at the content, I personally find it much more difficult to read through the resulting file.

rshuler: Maybe it can be made optional, but I would not want this to be the default.

eos-tech-support/eos-diagnostics
```diff
@@ -558,6 +558,11 @@ let diagnostics = [
         hardwareInfo: true,
         content: function() { return trySpawn('journalctl --no-pager --output verbose'); },
     },
+    {
+        title: 'Coredump',
```
rshuler: I think it is misleading to have this section if coredumps are not enabled.  Ideally, we would only include this section at all if coredumps are enabled.  So, an empty coredump list would actually mean that the core was not dumped.  Or, perhaps even better, always have the section but explicitly identify whether or not coredumps are enabled.

eos-tech-support/eos-diagnostics
```diff
@@ -673,6 +658,26 @@ function dumpHardwareInfo(filename) {
     return true;
 }
 
+function dumpHardwareInfo(filename) {
+    return dumpToArchive(filename, function(dir) {
+        copyAcpiFirmware(dir);
+
```
rshuler: Looks like the `copyCoredumps(dir);` line was omitted here.  Was that intentional?  If so, it seems that he commit message is incorrect that there are no functional changes.

pwithnall: Whoops, bad rebase. Fixed.

eos-tech-support/eos-diagnostics
pwithnall: Made it optional behind a `--verbose` flag (but it’s also enabled if running with `--coredumps`).

eos-tech-support/eos-diagnostics
pwithnall: Modified it to output a warning message and some stub content if `systemd-coredump` is not enabled in the kernel’s `core_pattern` sysctl.

pwithnall:
Updated branch to address all review comments.

eos-tech-support/eos-diagnostics
```diff
@@ -701,7 +706,7 @@ function dumpHardwareInfo(filename) {
     });
 }
 
-function dumpCoredumps(filename) {
+function dumpCoredumps(filename, verboseFlag) {
```
rshuler: I would not have `verboseFlag` as an input parameter to this function if it is going to be ignored and hard-coded.  Actually, I think it would be better to have the function respect this input parameter and instead force passing in true in the main routine when calling this function.

pwithnall: Fixed.

pwithnall:
Updated to address review comments, thanks.

PR #374, opened 2017-01-26
Fixes needed to generate ISOs from image builder

wjt:
https://phabricator.endlessm.com/T13559

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -256,3 +256,9 @@ Description: EndlessOS Technical Support Tools
  with the EndlessOS operating system.  They are not required as part of the
  normal user experience but are relatively small and helpful for advanced
  troubleshooting.
+
+Package: eos-tech-support-iso
+Architecture: all
+Depends: ${misc:Depends}, ${eos:Depends}
+Description: Create Endless OS ISO Images
+ Tools needed to create ISO images which are not needed in end-user images.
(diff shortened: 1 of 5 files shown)


debian/control
```diff
@@ -252,7 +252,13 @@ Package: eos-tech-support
 Architecture: all
 Depends: ${misc:Depends}, ${eos:Depends}
 Description: EndlessOS Technical Support Tools
- This is a collection technical support tools that are installed by default
+ This is a collection of technical support tools that are installed by default
  with the EndlessOS operating system.  They are not required as part of the
  normal user experience but are relatively small and helpful for advanced
  troubleshooting.
+
+Package: eos-tech-support-iso
```
wjt: This package doesn't actually get built, and I'm not sure why…

dbnicholson: So, the point here is that we want to pull the dependencies into the image builder but not into the ostree?

wjt: Exactly. We could of course choose to include these dependencies in the ostree – they're not *that* big – but I'd rather that question be separate from "can the image builder make ISOs".

dbnicholson: That's OK, I agree. It's not ideal (if we're shipping tools, we should actually provide the things they need to run), but in the short term this makes sense. I just wanted to make sure I understood the purpose of the separate metapackage.

PR #377, opened 2017-01-27
T14986: set the ISO volume ID and add autorun.inf

ramcq:
This adds a --label argument to eos-write-live-image in ISO mode, so that we can set more useful labels for the volume, visible under Windows, Mac and Linux. Also adds an autorun.inf to execute the installer automatically if the ISO is on an optical disk which is inserted on a Windows machine.

https://phabricator.endlessm.com/T14986

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -45,7 +45,7 @@ Options:
 
 ISO 9660-related options:
    -i,--iso                Write an ISO 9660 disk image to OUTPUT
-   -L,--label LABEL        ISO volume id
+   -L,--label LABEL        ISO volume id and Autorun.inf label
 
 Developer options (you probably don't want to use these):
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-write-live-image
```diff
@@ -373,6 +403,8 @@ if [ "$ISO" ]; then
     # Unzip the ISO bootzip content
     unzip -q -d "${DIR_IMAGES_ENDLESS}/grub/i386-pc/" "${BOOT_ZIP}" "iso/*"
 
+    VOLID=$(echo -n "$LABEL" | sed "s/[()]//g; s/[^[:alnum:]]/-/g")
```
wjt: Perhaps `sed "s/[^[:alnum:]]+/-/g; s/(^-|-$)//g"` which would collapse adjacent dashes too

ramcq: Achieved via tr.

PR #378, opened 2017-02-03
remove . from permitted strings in ISO volume ID

ramcq:
Removes "-volid text does not comply to ISO 9660 / ECMA 119 rules" error
from xorisso and should improve compatibility.

https://phabricator.endlessm.com/T14986


dbnicholson:
Whoops, accidentally put in request changes instead of approve there. Looks good.

PR #389, opened 2017-02-27
T15783

carlocaione:
https://phabricator.endlessm.com/T15783


eos-tech-support/eos-write-live-image
```diff
@@ -317,6 +317,7 @@ if [ "$ISO" ]; then
     ISO_TMPDIR=${TMPDIR:-/var/tmp}
     DIR_EFI="$(mktemp -d -t efi_dir.XXXXXX -p "$ISO_TMPDIR")"
     DIR_IMAGES="$(mktemp -d -t iso9660_dir.XXXXXX -p "$ISO_TMPDIR")"
+    DIR_ESP="$(mktemp -d -t esp_dir.XXXXXX -p "$ISO_TMPDIR")"
```
wjt: Maybe we could use:

```
ISO_TMPDIR="$(mktemp -d -t eos-iso.XXXXXX -p "${TMPDIR:-/var/tmp}")"
DIR_EFI="$ISO_TMPDIR/efi"
...
```

which simplifies the cleanup later? `$DIR_ESP` is only used to store a single file, after all.

carlocaione: added a new commit

PR #393, opened 2017-03-08
eos-tech-support: Port eos-updater service handling to latest unit files

pwithnall:
The eos-updater unit files changed recently (eos-updater.service was
created, having previously been an alias for eos-autoupdater.service).
Update the tech support scripts to reflect this.

Additionally, the configuration file locations for eos-updater changed:
/etc/eos-updater.conf moved to /etc/eos-updater/eos-autoupdater.conf.

https://phabricator.endlessm.com/T15977

### DIFF
--- a/eos-tech-support/eos-convert-system-qa
+++ b/eos-tech-support/eos-convert-system-qa
@@ -135,11 +135,10 @@ if $OSTREE || $APPS; then
         # Make sure the autoupdater doesn't start the updater after it's
         # killed below.
         echo "Stopping running eos-autoupdater."
-        systemctl stop eos-autoupdater.timer
-        systemctl stop eos-autoupdater.service
+        systemctl stop eos-autoupdater.timer eos-autoupdater.service
 
         echo "Killing eos-updater."
-        killall eos-updater &>/dev/null
+        systemctl stop eos-updater.service
 
         echo "Configuring OSTree for Staging."
         # HACK! HACK! HACK! ostree admin set-origin is horribly broken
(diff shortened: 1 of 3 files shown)


eos-tech-support/eos-convert-system-qa
```diff
@@ -154,8 +154,18 @@ if $OSTREE || $APPS; then
             --set=branches="${new_branch};"
 
         # Check for an update every time the updater runs
-        # rather than once every two weeks
-	sed -i 's/IntervalDays=14/IntervalDays=0/' /etc/eos-updater.conf
+        # rather than once every two weeks. eos-updater.conf is the old
+        # location of the config file.
+        if [ -f /etc/eos-updater.conf ]; then
```
rshuler: I don't understand the purpose of checking the old file location.  If the user has the new version of the script, it means that the user has already received an os upgrade, and thus will have the config file in the new location.  In fact, I think it might be possible (though I'm not sure) that the user will have an orphaned copy of the file in the old location, that will be completely ignored, in which case it would not be helpful to edit the old file rather than the new one.

pwithnall: If the user modified the old version of the config file before their latest OS upgrade, it will still be in place, I think? The three-way diff will delete the old version of the config file on upgrade if the user has not modified it.

In any case, `eos-autoupdater` reads its configuration from the old config file location if the new one doesn’t exist — the old one is not ignored.

wjt: But, post upgrade, won't the new location exist in its pristine state?

pwithnall: `/etc/eos-updater/eos-autoupdater.conf` will not exist; `/usr/share/eos-updater/eos-autoupdater.conf` will exist.

pwithnall: So, hmm, this code is not correct in the case that `/etc/eos-updater/eos-autoupdater.conf` and `/etc/eos-updater.conf` both exist, since `eos-autoupdater` loads the former (new location) in preference to the latter (old location). So running `sed` on the old location won’t help.

What we want to do is:
 1. If the new location exists, patch that.
 2. Otherwise, if the old location exists, patch that.
 3. Otherwise, copy the defaults to the new location and patch that.

pwithnall: Patch updated to do this.

eos-tech-support/eos-convert-system-qa
```diff
@@ -154,8 +153,17 @@ if $OSTREE || $APPS; then
...
 
         # Check for an update every time the updater runs
-        # rather than once every two weeks
-	sed -i 's/IntervalDays=14/IntervalDays=0/' /etc/eos-updater.conf
+        # rather than once every two weeks. eos-updater.conf is the old
+        # location of the config file.
+        sed -i 's/IntervalDays=14/IntervalDays=0/' /etc/eos-updater/eos-autoupdater.conf 2>/dev/null || true
+        sed -i 's/IntervalDays=14/IntervalDays=0/' /etc/eos-updater.conf 2>/dev/null || true
+
+        if [ ! -f /etc/eos-updater/eos-autoupdater.conf ] && [ ! -f /etc/eos-updater.conf ]; then
```
wjt: Under what circumstances would neither file exist, besides "user goes out of their way to delete them"?

pwithnall: This is (will be) a fairly common case: the case where a new image is installed — `/etc/eos-updater.conf` would not exist because it’s old/deprecated, and `/etc/eos-updater/*` would not exist because they need to be manually copied and edited from the defaults in `/usr/share/eos-updater/`.

PR #395, opened 2017-03-10
eos-tech-support: Add eos-deploy-debs script

pwithnall:
This deploys one or more .deb archives to a non-dev-converted system
over SSH, putting them in a temporary OSTree overlay which is lost on
reboot.

https://phabricator.endlessm.com/T16088

### DIFF
--- a/eos-tech-support/eos-deploy-debs
+++ b/eos-tech-support/eos-deploy-debs
@@ -1,0 +1,140 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+
+# Copyright © 2017 Endless Mobile, Inc.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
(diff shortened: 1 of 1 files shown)


pwithnall:
There is little point to it being on the image; it just needs to be somewhere where we can all get at it, most likely as a symlink from a git clone into `$PATH`.

wjt:
Everything in the `eos-tech-support` directory ends up in the image.

pwithnall:
> Everything in the eos-tech-support directory ends up in the image.

Got a better suggestion of where to put it? Is there already a directory somewhere of scripts which don’t get installed in an image, but are useful for people to use?

wjt:
Sorry, when I suggested this package on Slack I assumed you wanted it in the image. Possibly somewhere in `eos-build` though I'm never quite sure of the boundaries there between "stuff run by jenkins and OBS" and "random scripts"…

rshuler:
We have a couple repos that might make sense... eos-customer-support (public) or eos_private-support (private).

rshuler:
And there might be another place if it is just for developers.

pwithnall:
> And there might be another place if it is just for developers.

It is targeted at developers. Would `eos-build` be the most appropriate in that case then?

rshuler:
I would not use eos-build for things that would be run on the OS.  The repo I was thinking about is eos-dev-config.

But looking at the comments, if this is something that would be run on a non-converted system, I think including it in the deployed images as part of eos-meta/eos-tech-support makes sense.

pwithnall:
> I would not use eos-build for things that would be run on the OS. The repo I was thinking about is eos-dev-config.

eos-dev-config looks like the best fit so far, since this script targets similar use cases as the ones in eos-dev-config/bin.

 * It does not need to be in the image
 * It is targeted at developers
 * It needs to be run on the developer’s machine (which might be EOS), not the EOS system they want to deploy packages to

In all of those senses, it’s pretty similar to `eos-dev-config/bin/raw_img_to_vdi`. I’ll close this PR and open a new one against eos-dev-config.

pwithnall:
⇒ https://github.com/endlessm/eos-dev-config/pull/59

PR #397, opened 2017-03-16
Update eos-tech-support tools to handle ISOs

wjt:
I am *more* than happy to defer the final commit to a separate PR. It looks like it rewrites the file, but it's mostly indenting 4 spaces to the right and replacing repeated arguments with `self.foo`. Much easier to read in `git show --ignore-space-change`.

https://phabricator.endlessm.com/T16171

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -298,7 +298,7 @@ class EosDownloadImage(object):
         '''Downloads a single OS image, and its signatures. If args.iso is True,
         fetches the ISO; if not, fetches the full raw image, plus the
         corresponding bootloader bundle.'''
-        _manifest, images = self.fetch_manifest()
+        _, images = self.fetch_manifest()
         args = self.args
 
         if args.version is not None:
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -56,310 +56,306 @@ else:
         pass
 
 
-def download(session, src, dest, src_size=None, src_mtime=None):
-    """Download a file using a Requests Session
+class EosDownloadImage(object):
+    def download(self, src, dest, src_size=None, src_mtime=None):
+        """Download a file using a Requests Session
 
-    This acts like wget --continue or rsync where the sizes and
-    modification times of the source and destination files are
-    considered.
+        This acts like wget --continue or rsync where the sizes and
+        modification times of the source and destination files are
+        considered.
 
-    This handles the case where the file is larger than the CDN allows
-    by using a provided size and splitting up the requests across
-    ranges.
-    """
-    # By default, requests sets Accept-Encoding to gzip, deflate, but we
-    # want raw bytes
-    headers = requests.utils.default_headers()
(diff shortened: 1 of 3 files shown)


eos-tech-support/eos-download-image
```diff
@@ -315,12 +320,21 @@ def fetch_image(session, args, base):
...
-    if 'boot' in personality and personality['boot']['file']:
-        download_and_verify(session, base, personality['boot'],
-                            what='boot', destdir=args.outdir)
+    if args.iso:
+        what = 'iso'
+    else:
+        what = 'full'
+        if 'boot' in personality and personality['boot']['file']:
+            download_and_verify(session, base, personality['boot'],
+                                what='boot', destdir=args.outdir)
```
dbnicholson: Just noticed this now, but all the other calls to `download_and_verify` are wrapped in `print` except this one.

eos-tech-support/eos-download-image
```diff
@@ -56,310 +56,306 @@ else:
...
+
+        with requests.sessions.session() as session:
+            self.session = session
+
+            if args.windows_tool:
+                self.fetch_windows_tool()
+            elif args.mirror:
+                self.mirror_images()
+            else:
+                self.fetch_image()
```
dbnicholson: This block feels awkward in a class, but I guess it's more of a one time object than a class that you instantiate and work with.

eos-tech-support/eos-write-image
```diff
@@ -120,10 +121,11 @@ if ! which pv >/dev/null 2>&1; then
 fi
 
 cat_image() {
-    if file --mime-type "$IMAGE" | grep -q xz$; then
+    MIME_TYPE="$(file --mime-type "$IMAGE")"
+    if echo "$MIME_TYPE" | grep -q xz$; then
```
wjt: Much nicer!

eos-tech-support/eos-download-image
```diff
@@ -282,63 +284,68 @@ def fetch_windows_tool(session, args, base):
...
     images = list(manifest['images'].values())
     images.sort(key=lambda i: distutils.version.LooseVersion(i['version']))
-    if args.mirror:
-        selected = images
-    elif args.version is not None:
+    return manifest, images
+
+
+def fetch_image(session, args, base):
+    _manifest, images = fetch_manifest(session, args, base)
```
wjt: It does look neater this way, so I've made this change. It's functionally equivalent though:
```python
>>> _, y = (1, 2)
>>> print(_)
1
```

eos-tech-support/eos-download-image
wjt: This is deliberate. The script emits the path to the downloaded image file on stdout, which is consumed by `eos-write-live-image`. I'll add a comment because this is not obvious.

eos-tech-support/eos-download-image
wjt: Yes. I suppose it would be cleaner to have `__main__` go something like:

```python
with requests.session() as session:
    app = EosDownloadImage(session)
    app.run()
```

PR #400, opened 2017-04-03
eos-download-image: print path to Windows tool

wjt:
This is a regression introduced in 42c0d01. eos-write-live-image expects
this script to print the path to the newly-downloaded file to stdout,
when called with --latest.

Without this fix, `eos-write-live-image` without an explicit
--windows-tool argument will fail.

https://phabricator.endlessm.com/T16099

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -284,6 +284,7 @@ class EosDownloadImage(object):
         dest = os.path.join(self.args.outdir, filename)
         log("Fetching Windows Installer")
         self.download(url, dest)
+        print(dest)
 
     def fetch_manifest(self):
         manifest_url = '{base}/releases-{product}-3.json'.format(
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-download-image
```diff
@@ -284,6 +284,7 @@ class EosDownloadImage(object):
         dest = os.path.join(self.args.outdir, filename)
         log("Fetching Windows Installer")
         self.download(url, dest)
+        print(dest)
```
dbnicholson: Could you add a comment on why this is needed? Either in the function docstring or just a normal comment above the print?

wjt: I've gone a little belt-and-braces, and documented it both in docstrings, and in `--help`. Happy to remove one or the other if you prefer, but now that you mention it, it does seem worth mentioning in `--help` as well, since it's part of the "API" of this tool.

PR #406, opened 2017-04-13
eos-add-to-desktop: Display usage for -h, --help

manuq:
And display the basename of the script in the usage, not the entire
path.


eos-tech-support/eos-add-to-desktop
```diff
@@ -4,9 +4,9 @@
 # The appname.dekstop file must already exist (typically in
```
wjt: dekstop!

PR #411, opened 2017-05-12
tech support: add eos-revert-ostree

rshuler:
This script reverts to the previously deployed OSTree,
in the case that the new OSTree can be booted but has a
critical problem that makes it unusable (e.g., broken WiFi).

https://phabricator.endlessm.com/T16972

### DIFF
--- a/eos-tech-support/eos-rollback
+++ b/eos-tech-support/eos-rollback
@@ -1,0 +1,43 @@
+#!/bin/bash -e
+
+# Reverts to the previously deployed OSTree
+
+# This assumes that there are exactly two OSTrees available:
+# - the currently deployed  (listed as "* eos <commit>" in the status)
+# - the previously deployed (listed as "  eos <commit>" in the status)
+#
+# This may take several minutes to traverse the entire directory
+# tree to check for file objects to prune.  While not ideal
+# (since we really just want to switch the boot order
+# without a need to check for pruning), this is the best we can do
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-revert-ostree
```diff
@@ -0,0 +1,43 @@
...
+
+echo
+read -p "Are you sure you want to revert to the previously deployed OSTree? [y/N] "
+response=${REPLY,,} # to lower
+if [[ ! $response =~ ^(yes|y)$ ]]; then
+    exit 1
+fi
+echo
+
+deploy=`ostree admin status | grep '  eos' | awk '{print $2}'`
```
dbnicholson: Since `awk` also operates on regular expressions, you can use it to match the expression. You add a `/<regex>/` before the expression. Furthermore, I think you want to anchor the regex with `^` so you don't accidentally match an `eos` somewhere else in the output. Finally, it would be unusual for this to happen, but I think you should guard against having more than one old deployment. Putting it all together:
```
deploy=$(ostree admin status | awk '/^  eos /{print $2}' | head -n1)
```
It's unfortunate that we have to scrape the `ostree admin status` output like this.

rshuler: Thank you!  That is indeed better.

rshuler:
Yes, I like eos-rollback.  Thanks for the suggestion.

PR #414, opened 2017-05-24
ISO: embed md5sums

wjt:
Untested! And non-working. Interestingly the MD5 sums embedded by this flag – if indeed there are any – are not the same as those embedded by `implantisomd5`.


wjt:
Obsolete since ISO creation has moved to the image builder.

PR #424, opened 2017-06-17
T17676

rshuler:
https://phabricator.endlessm.com/T17676

### DIFF
--- a/eos-tech-support/eos-convert-system-qa
+++ b/eos-tech-support/eos-convert-system-qa
@@ -1,10 +1,5 @@
 #!/bin/bash -e
 
-# OSTree dev repo URL settings.
-OSTREE_USER=endless
-OSTREE_HOST=origin.ostree.endlessm.com
-OSTREE_DEV_ROOT=staging/dev
-
 function usage {
     cat <<EOF
 Invalid option: $1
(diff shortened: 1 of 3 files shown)


eos-tech-support/eos-convert-system-qa
```diff
@@ -91,79 +86,14 @@ if $OSTREE || $APPS; then
...
-        echo "Setting flatpak runtimes URL to $new_runtime_url"
-        flatpak remote-modify eos-runtimes --url="$new_runtime_url"
-
-        apps_url=$(ostree --repo="$flatpak_repo" config get 'remote "eos-apps".url')
-        apps_repo=${apps_url##*/}
-        new_apps_url="https://${OSTREE_USER}:${OSTREE_PASSWORD}@${OSTREE_HOST}/${OSTREE_DEV_ROOT}/${apps_repo}"
-
-        echo "Setting flatpak apps URL to $new_apps_url"
-        flatpak remote-modify eos-apps --url="$new_apps_url"
+	eos-stage-flatpak dev "${OSTREE_PASSWORD}"
```
dbnicholson: Please expand this tab to spaces.

eos-tech-support/eos-convert-system-qa
```diff
@@ -91,79 +86,14 @@ if $OSTREE || $APPS; then
...
-        sed -i 's/IntervalDays=14/IntervalDays=0/' /etc/eos-updater/eos-autoupdater.conf 2>/dev/null || true
-        sed -i 's/IntervalDays=14/IntervalDays=0/' /etc/eos-updater.conf 2>/dev/null || true
-
-        if [ ! -f /etc/eos-updater/eos-autoupdater.conf ] && [ ! -f /etc/eos-updater.conf ]; then
-            mkdir -p /etc/eos-updater
-            cp /usr/share/eos-updater/eos-autoupdater.conf /etc/eos-updater
-
-            sed -i 's/IntervalDays=14/IntervalDays=0/' /etc/eos-updater/eos-autoupdater.conf
-        fi
+	eos-stage-ostree dev "${OSTREE_PASSWORD}"
```
dbnicholson: Same tabs vs spaces issue.

eos-tech-support/eos-convert-system-qa
dbnicholson: Tabs vs spaces.

eos-tech-support/eos-convert-system-qa
dbnicholson: Tabs vs spaces.

eos-tech-support/eos-stage-flatpak
```diff
@@ -0,0 +1,72 @@
...
+    PASSWORD="$2"
+fi
+
+if [ $EUID != 0 ] ; then
+    echo "Program requires superuser privileges" >&2
+    exit 1
+fi
+
+if [ "$STAGE" == "dev" ] ; then
+    if [ $# -lt 2 ] ; then
```
dbnicholson: I think it would be better to check `[ -z "$PASSWORD" ]` here since you already handled the argument. That's more future proof if the order of arguments change or you `shift` any arguments.

eos-tech-support/eos-stage-ostree
```diff
@@ -0,0 +1,112 @@
...
+    PASSWORD="$2"
+fi
+
+if [ $EUID != 0 ] ; then
+    echo "Program requires superuser privileges" >&2
+    exit 1
+fi
+
+if [ "$STAGE" == "dev" ] ; then
+    if [ $# -lt 2 ] ; then
```
dbnicholson: Same comment about checking `$PASSWORD` instead of the number of arguments.

eos-tech-support/eos-stage-ostree
```diff
@@ -0,0 +1,112 @@
...
+
+# Get the current URL and convert to the new stage.
+url=$(ostree config get 'remote "eos".url')
+repo=${url##*/}
+new_url="${base_url}/${repo}"
+
+# Make sure the autoupdater doesn't start the updater after it's
+# killed below.
+echo "Stopping running eos-autoupdater."
+systemctl stop eos-autoupdater.timer eos-autoupdater.service
```
dbnicholson: I don't think you want to stop automatic updates when the target stage is demo or prod. If you're telling general users to run this script to get on the demo release, then they probably won't know the details to restart the auto updater.

Probably you can just add a section at the end:
```
if [ "$STAGE" = demo ] || [ "$STAGE" = prod ]; then
   systemctl start eos-autoupdater.timer
fi
```
Possibly you could start `eos-autoupdater.service` as well so the upgrade starts right away. On the other hand, if the demo commit is older than their current commit, that won't go anywhere. So, they'd probably need an explicit `ostree admin upgrade --allow-downgrade`.

rshuler: The instructions I would provide would be to run this script and then manually check for updates from the settings detail page.  I'd prefer not to allow downgrades by default, as one use case is to download the latest eos3.2 staging image and configure it so that it will get the next beta release (but stick on the current staging in the meantime).  Also, if a user wants to switch back from demo to prod, I would not want to automatically downgrade the user immediately -- I'd prefer that the normal case is that they don't get any more beta updates, but they wait where they are for the next prod release.  If anyone really wants a downgrade, we can provide explicit instructions.

That said, if the user does not immediately check for the updates, I think it does make sense to re-enable the autoupdater timer.  And, I think it is fine to do that on dev, too.

So, unless you object, I'm just going to turn the autoupdater back on at the end, regardless of stage.

rshuler: Actually, I like the idea of starting the eos-autoupdater.service right away, too, so I don't have to instruct the user to manually check for updates.  My only real concern was automatically downgrading.

dbnicholson: Except that breaks QA mode in my opinion. When I'm trying to test, the first thing I do is turn off the autoupdater so that the system stays in that state. That's why it's turned off in the first place.

dbnicholson: On https://phabricator.endlessm.com/T17676 you said you got @IzzyC's agreement to do this, so I'm OK to go ahead. We can always remove it if it's problematic for QA.

eos-tech-support/eos-stage-ostree
```diff
@@ -0,0 +1,120 @@
...
+    cp /usr/share/eos-updater/eos-autoupdater.conf /etc/eos-updater
+
+    sed -i "s/IntervalDays=.*/IntervalDays=$interval_days/" /etc/eos-updater/eos-autoupdater.conf
+fi
+
+# Restart the autoupdater timer, and try to update now.
+# Note that downgrades (e.g., from demo to previous prod release)
+# won't happen automatically.  They would have to be done via
+# `ostree admin upgrade --allow-downgrade`, but that should be
+# a manual step if someone really wants to downgrade.
```
dbnicholson: Note that the "downgrade" may not really be a downgrade if you're switching stages since "downgrade" just means the commit is older than what you have. If you switch from dev to demo and the demo commit happened to have been made before the dev commit (likely since we rarely promote dev to demo), then there will be no update.

PR #425, opened 2017-06-26
Add OpenVPN plugin for network-manager

rshuler:
https://phabricator.endlessm.com/T17912

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -8,6 +8,7 @@ alsa-utils
 apt
 apt-transport-https
 apt-utils
+bluez-obexd
 brasero
 btrfs-tools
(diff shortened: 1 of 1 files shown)


eos-core-depends
```diff
@@ -112,6 +112,7 @@ nautilus
 bluez-obexd
 openprinting-ppds
 openssh-server
+openvpn
```
cosimoc: I think you want `network-manager-openvpn`, no? (We may not have the package built on OBS)

rshuler: Okay, makes sense.  We have that in OBS, too, and it pulls in openvpn as a dependency.

PR #426, opened 2017-06-30
Add previously-recommended packages to core

dsd:
We are moving to not installing recommended packages in the ostree
builder.

However, we do ship some important packages via the Recommends
mechanism. Add explicit dependencies on these packages here, so that
they continue to be shipped.

https://phabricator.endlessm.com/T5124


dsd:
rebased on latest master

PR #432, opened 2017-07-27
eos-diagnostics: update for /var/lib/dpkg presence

wjt:
This matches f934324 which made the same change for eos-dev-fix.

https://phabricator.endlessm.com/T18503

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -396,8 +396,8 @@ function collectChromiumPluginsInfo() {
 
 function collectEosDevInfo() {
     let output = '';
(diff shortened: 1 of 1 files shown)


dsd:
This change made sense for eos-dev-fix because eos-dev-fix creates /var/lib/xml-core.
But I don't see this as clearly in eos-convert-system. Does eos-convert-system definitely create this which we can use as  a flag here?

ramcq:
eos-convert-system reconstitutes /var like it is on a Debian system, copying back our archived copy from /usr/var. That's what used to put /var/lib/dpkg back - arguably, this is wasted space in the ostree and we should not ship any of it, now that /var/lib/dpkg is in a more useful place.

PR #433, opened 2017-07-31
eos-rollback: Rollback even if the old deployment is booted

rshuler:
Rather than switch to the deployment that is not currently booted,
we change the default boot order.  This way, if the user has to
go into the GRUB menu to boot into the previous deployment,
he/she can still use eos-rollback to persist this old deployment.

Each call to this script rotates the boot order for the available
deployments, putting the first deployment to the end of the list.
In the case of Endless OS, there are only two deployments,
so we effectively toggle between the two available versions.

The implementation is migrated from awful command-line parsing
in bash to a clean python script provided by Dan Nicholson.

With the python API, we no longer have to redeploy, so the process
is much faster.

https://phabricator.endlessm.com/T18539

### DIFF
--- a/eos-tech-support/eos-rollback
+++ b/eos-tech-support/eos-rollback
@@ -1,48 +1,92 @@
-#!/bin/bash -e
+#!/usr/bin/python3 -u
 
-# Reverts to the previously deployed OSTree
+# Changes the boot order to revert to the previous OSTree deployment.
 
-# This assumes that there are exactly two OSTrees available:
-# - the currently deployed  (listed as "* eos <commit>" in the status)
-# - the previously deployed (listed as "  eos <commit>" in the status)
-#
-# This may take several minutes to traverse the entire directory
-# tree to check for file objects to prune.  While not ideal
-# (since we really just want to switch the boot order
-# without a need to check for pruning), this is the best we can do
-# with the curently available command line options.
-#
-# Note the importance of specifying the contents of the origin file.
-# Otherwise, the origin file hard codes the commit hash,
-# which prevents future upgrades.
+# On Endless OS, after the first OS upgrade, we keep two deployments available.
+# So, each call to this script will toggle between the two deployments.
 
-USERID=$(id -u)
-if [ "$USERID" != "0" ]; then
-    echo "Program requires superuser privileges"
-    exit 1
-fi
+from argparse import ArgumentParser
+import gi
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-rollback
```diff
@@ -1,48 +1,92 @@
...
+
+# Changes the boot order to revert to the previous OSTree deployment.
+
+# On Endless OS, after the first OS upgrade, we keep two deployments available.
+# So, each call to this script will toggle between the two deployments.
+
+from argparse import ArgumentParser
+import gi
+gi.require_version('OSTree', '1.0')
+from gi.repository import GLib, Gio, GObject, OSTree
```
dbnicholson: You can drop GLib and GObject from here. I don't know why I left those in there.

eos-tech-support/eos-rollback
```diff
@@ -1,48 +1,92 @@
...
+import os
+import sys
+import time
+
+if os.getuid() != 0:
+    print('Program requires superuser privileges')
+    exit(1)
```
dbnicholson: Apparently, you're only supposed to used `exit` from the interactive shell - https://docs.python.org/3/library/constants.html?highlight=exit#exit. Instead, you're supposed to use `sys.exit` - https://docs.python.org/3/library/sys.html?highlight=sys.exit#sys.exit. It's not necessary to change, but it would be nice to clean those up here.

eos-tech-support/eos-rollback
```diff
@@ -1,48 +1,92 @@
...
+
+from argparse import ArgumentParser
+import gi
+gi.require_version('OSTree', '1.0')
+from gi.repository import GLib, Gio, GObject, OSTree
+import os
+import sys
+import time
+
+if os.getuid() != 0:
```
dbnicholson: Might want to delay this check until after the arguments have been parsed so you can at least see the help text when you're unprivileged.

eos-tech-support/eos-rollback
```diff
@@ -1,48 +1,92 @@
...
+    print('Program requires superuser privileges')
+    exit(1)
+
+print()
+response = input('Are you sure you want to rollback to the previous OS version? [y/N] ')
+response = response.lower()
+if response != 'yes' and response != 'y':
+    exit(1)
```
dbnicholson: Same with this block going after `parse_args()`.

PR #434, opened 2017-08-02
eos-core: eos-google-chrome-helper was renamed

cosimoc:
https://phabricator.endlessm.com/T16839

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -30,7 +30,7 @@ eos-discovery-feed
 eos-event-recorder-daemon
 eos-exploration-center
 eos-gates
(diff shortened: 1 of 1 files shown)


eos-core-depends
```diff
@@ -30,7 +30,7 @@ eos-discovery-feed
 eos-event-recorder-daemon
 eos-exploration-center
 eos-gates
-eos-google-chrome-helper [amd64]
+eos-install-app-helper [amd64]
```
rshuler: Drive-by comment: I suppose it is appropriate for now that this is only amd64, but in the long run, I suspect we will have some ARM apps in this state.

PR #443, opened 2017-09-20
eos-write-live-image: remove --iso, cleanup

wjt:
ISO generation now lives in the image builder.  It was never possible to run this mode on an unconverted Endless OS system, since several of the dependencies are not included in the ostree.

I also tidied up a few things I noticed along the way.

https://phabricator.endlessm.com/T17203

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -439,7 +439,7 @@ if [ ! "$WRITABLE" ]; then
     echo "$OS_IMAGE_UNCOMPRESSED_BASENAME" > "${DIR_IMAGES_ENDLESS}/live"
     cp "$WINDOWS_TOOL" "$DIR_IMAGES/"
     WINDOWS_TOOL_BASENAME="$(basename "$WINDOWS_TOOL")"
-    iconv -f utf-8 -t utf-16 <<AUTORUN_INF > "${DIR_IMAGES}/autorun.inf"
+    sed 's/$/\r/' <<AUTORUN_INF | iconv -f utf-8 -t utf-16 > "${DIR_IMAGES}/autorun.inf"
 [AutoRun]
 label=${LABEL}
 icon=${WINDOWS_TOOL_BASENAME}
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -450,24 +450,6 @@ MusicFiles=false
 PictureFiles=false
 VideoFiles=false
 AUTORUN_INF
-
-    # Work around a bug in eos-installer, which only shows images whose
-    # personality is 'base' or a valid locale. Delete this block once
-    # 3.1.1 is released.
-    if [[ "${OS_IMAGE_UNCOMPRESSED_BASENAME}" == fnde-eos3.1-*.17011*.fnde_* ]]; then
-       for i in "$DIR_IMAGES_ENDLESS"/*; do
-          j=${i/fnde_aluno/base}
-          j=${j/fnde_escola/pt_BR}
-          if [ "$i" != "$j" ]; then
-             mv $i $j
-          fi
-       done
-
-       j=${OS_IMAGE_UNCOMPRESSED_BASENAME}
-       j=${j/fnde_aluno/base}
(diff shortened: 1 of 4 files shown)


eos-tech-support/eos-write-live-image
```diff
@@ -45,7 +45,8 @@ LABEL=
 DEBUG=
 
 usage() {
-    local SELF="$(basename "$0")"
+    local SELF
+    SELF=$(basename "$0")
```
dbnicholson: Hmm, that's an interesting one. I wonder what was wrong with the original.

eos-tech-support/eos-write-live-image
```diff
@@ -174,7 +175,7 @@ if [ $# -ne 1 ] ; then
     if [ $# -lt 1 ] ; then
         echo "Error: missing OUTPUT" >&2
     else
-        echo "Error: extra arguments after OUTPUT: $@" >&2
+        echo "Error: extra arguments after OUTPUT:" "$@" >&2
```
dbnicholson: What does this do?

eos-tech-support/eos-write-live-image
wjt: https://gist.github.com/wjt/2bcb4a9705db683c8d8292e3cc4ac9d3

```
  local x=$(echo oh no; false)
        ^-- SC2155: Declare and assign separately to avoid masking return values.
```

wjt: Also the outer quotes are redundant, I think

eos-tech-support/eos-write-live-image
wjt:
```
echo "Arguments: $@"
                 ^-- SC2145: Argument mixes string and array. Use * or separate argument.
```

In this case there's not much in it, given this is just error output – I picked one arbitrarily.

eos-tech-support/eos-write-live-image
dbnicholson: Interesting. Ah, this is (as the warning says) when you're mixing declarations (`local`, `export`, etc) with assignment. Good to know!
```
$ x=$(false)
$ echo $?
1
$ export x=$(false)
$ echo $?
0
$ f() { local x=$(false); echo $?; }
$ f
0
```
I always leave the quotes out when doing variable assignment with command substitution, but only because I've experienced it to be unnecessary in bash and dash. I'm not sure why that's the case, though.

eos-tech-support/eos-write-live-image
dbnicholson: https://github.com/koalaman/shellcheck/wiki/SC2145

I'd never thought about that. Here we're just passing to `echo`, so who cares how the arguments are split. But, I can see this being interesting in some other cases depending on exactly what you're trying to do with `$@` or some other array.
```
$ set -- "a b" c
$ a=("args: $@")
$ declare -p a
declare -a a='([0]="args: a b" [1]="c")'
$ a=("args: $*")
$ declare -p a
declare -a a='([0]="args: a b c")'
$ a=("args:" "$@")
$ declare -p a
declare -a a='([0]="args:" [1]="a b" [2]="c")'
```

eos-tech-support/eos-write-live-image
wjt: `man dash` does not have anything much to say on this. `man bash` says under `PARAMETERS` (emphasis mine):

> A variable may be assigned to by a statement of the form
>
> ```bash
> name=[value]
> ```
>
> If  value is not given, the variable is assigned the null string.  All values undergo tilde expansion, parameter and variable expansion, command substitution, arithmetic expansion, and quote removal (see EXPANSION below).   If  the  variable  has  its integer  attribute  set,  then value is evaluated as an arithmetic expression even if the $((...)) expansion is not used (see Arithmetic Expansion below).  **Word splitting is not performed**, with the exception of "$@" as explained  below  under  Special Parameters.

eos-tech-support/eos-write-live-image
wjt: Interesting! I find bash's array handling deeply mysterious at the best of times, and definitely hadn't thought about this case.

PR #444, opened 2017-09-22
convert-system-qa: set core_pattern immediately

wjt:
Previously, systemd-coredump would be enabled on the next boot, but not
during the current boot. I hit an eos-updater SIGSEGV while pulling an
update after running `eos-convert-system-qa --ostree` and it would have
been nice to get a core dump.


dbnicholson:
Good call.

PR #445, opened 2017-09-26
ostree-fix-repo: Improve check for partial commits

dbnicholson:
Unfortunately, the script was only catching when a `dirtree` object was missing, not a `dirmeta` or `file` object. That meant that in many cases the commit wasn't being marked as partial and the script carried on without refetching missing objects.

I also added one more commit to skip pulling full Locale commits as flatpak intentionally does partial pulls of those. Pulling the full commit would just waste bandwidth and disk space.

I tested this in a VM where I deleted all types of missing objects. After running the script, `ostree fsck` succeeded, I was able to update to update the platform, and I was able to install an app.

https://phabricator.endlessm.com/T18946

### DIFF
--- a/eos-tech-support/eos-fix-ostree-repo
+++ b/eos-tech-support/eos-fix-ostree-repo
@@ -201,17 +201,18 @@ def pull_partial_refs(repo):
     # Look for any partial refs and re-pull them.
     _, all_refs = repo.list_refs()
     for refspec, checksum in all_refs.items():
+        _, remote, ref = OSTree.parse_refspec(refspec)
+        if remote is None:
+            # Don't bother pulling local refs. Only the ostree deploys
+            # are local, and as long as they're marked partial, they can
+            # be updated later.
+            continue
+
         _, commit, state = repo.load_commit(checksum)
         if state != OSTree.RepoCommitState.REPO_COMMIT_STATE_PARTIAL:
             continue
 
         # Try to pull the full commit again.
-        _, remote, ref = OSTree.parse_refspec(refspec)
-        if remote is None:
-            # If there's no remote, assume it's an ostree ref and use
-            # "eos" as the remote.
-            print('No remote for ref', ref, 'assuming "eos"')
-            remote = 'eos'
         print('Pulling', checksum, 'commit from', remote, 'for', ref)
         pull_commit(repo, remote, checksum, full=True)
 
--- a/eos-tech-support/eos-fix-ostree-repo
+++ b/eos-tech-support/eos-fix-ostree-repo
@@ -166,13 +166,29 @@ def mark_commits_partial(repo):
             print('Commit', checksum, 'already marked as partial')
             continue
 
+        mark_partial = False
         try:
-            repo.traverse_commit(checksum, 0)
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-fix-ostree-repo
```diff
@@ -151,12 +152,20 @@ def fix_dangling_refs(repo):
             pull_commit(repo, remote, checksum)
 
 
-def mark_commits_partial(repo):
+def get_all_objects(repo):
+    """Get a dictionary of all objects in the repository"""
+    repo_path = os.path.realpath(repo.get_path().get_path())
+    print('Finding all objects in', repo_path)
+
+    return repo.list_objects(OSTree.RepoListObjectsFlags.ALL, None)[1]
```
pwithnall: Nitpick: Would maybe be a bit clearer to use tuple matching on the return values:
```
_, all_objects = repo.list_objects(OSTree.RepoListObjectsFlags.ALL, None)
return all_objects
```

eos-tech-support/eos-fix-ostree-repo
```diff
@@ -166,29 +175,55 @@ def mark_commits_partial(repo):
...
 
+        mark_partial = False
         try:
-            repo.traverse_commit(checksum, 0)
+            # If a dirtree is missing, traverse_commit will fail with
+            # G_IO_ERROR_NOT_FOUND.
+            _, reachable_objects = repo.traverse_commit(checksum, 0)
+
+            # Unfortunately, it doesn't check that the leaves (dirmeta
+            # and files) exist, so we need to do that manually. In case
```
pwithnall: This sounds like something which it should do. Have you filed an upstream issue about `traverse_commit()`’s behaviour? Maybe do that and link to it in a `FIXME` comment here.

eos-tech-support/eos-fix-ostree-repo
dbnicholson: Sure. I had that first then decided it was noisy :)

eos-tech-support/eos-fix-ostree-repo
dbnicholson: https://github.com/ostreedev/ostree/issues/1222. Will add a comment.

eos-tech-support/eos-fix-ostree-repo
```diff
@@ -166,29 +175,55 @@ def mark_commits_partial(repo):
...
+            for commit_obj in reachable_objects:
+                if commit_obj in found_objects:
+                    # Already found this object
+                    continue
+
+                commit_obj_checksum, commit_obj_type = \
+                    OSTree.object_name_deserialize(commit_obj)
+
+                _, have_object = repo.has_object(commit_obj_type,
+                                                 commit_obj_checksum)
```
dbnicholson: Oh, I just realized I don't have to do this lookup at all. I already passed in the set of all objects in the repo in `all_objects`. I'm going to test that out now.

dbnicholson:
CR ready again. @pwithnall sorry, this is a bit of a redo on this PR although the concepts are the same.

PR #446, opened 2017-09-29
eos-fix-ostree-repo workarounds for old OSTree

dbnicholson:
The ostree on eos3.0 has a couple issues that make the script fail. Add some workarounds for them and a message at the end so they know if it worked or not.

https://phabricator.endlessm.com/T19420

### DIFF
--- a/eos-tech-support/eos-fix-ostree-repo
+++ b/eos-tech-support/eos-fix-ostree-repo
@@ -355,6 +355,8 @@ def main():
     # so there are no longer any missing objects
     pull_partial_refs(repo)
 
+    print('\nSuccess! Try to update the OS and Apps now.')
+
 
 if __name__ == '__main__':
     main()
--- a/eos-tech-support/eos-fix-ostree-repo
+++ b/eos-tech-support/eos-fix-ostree-repo
@@ -330,6 +330,21 @@ def main():
               file=sys.stderr)
         sys.exit(1)
 
+    # In older OSTree, cleaning up after a transaction (e.g., a pull)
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-fix-ostree-repo
```diff
@@ -52,6 +52,77 @@ import sys
...
+                    # Not a partial sha256
+                    continue
+
+                if ext == '':
+                    continue
+                elif (ext == '.filez' and
+                      repo_mode == OSTree.RepoMode.ARCHIVE_Z2):
+                    objtype = OSTree.ObjectType.FILE
+                elif (ext == '.file' and
+                      repo_mode != OSTree.RepoMode.ARCHIVE_Z2):
```
pwithnall: Technically that should be `and repo_mode in (OSTree.RepoMode.BARE, .BARE_USER, .BARE_USER_ONLY)`. The expressions are equivalent for the moment, but they might change over time.

Then again, some of those `RepoMode`s probably didn’t exist in the version of OSTree in EOS 3.0. Sigh. Probably best to leave it as you’ve written it then.

dbnicholson: Yeah. I could make a copy of ostree's internal `mode_is_bare` helper and only match on whatever's in the `RepoMode` enum besides `ARCHIVE_Z2`. Hmm, that looks kind of ugly as I think you have to look at the internal `__enum_values__` property.

pwithnall: Exactly. I think it’s probably best left as you’ve written it.

PR #455, opened 2017-11-21
eos-download-image: Allow download from URL

mwleeds:
For Solutions images we don't want to publish an index of them all, but
we still need to be able to download them using e-d-i. So add --url and
--size arguments so the process of checking a manifest can be
circumvented.

https://phabricator.endlessm.com/T20033

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -177,30 +177,15 @@ class EosDownloadImage(object):
 
         return keyring
 
-    def download_and_verify(self, meta, what, url=None, size=None):
-        """ Download and verify image. Caller should either use meta or specify url and size."""
+    def download_and_verify(self, meta, what):
         destdir = self.args.outdir
-        if meta is not None:
-            image_url = '{}/{}'.format(self.base, meta['file'])
-            image_size = meta['compressed_size']
+        image_url = '{}/{}'.format(self.base, meta['file'])
+        image_size = meta['compressed_size']
+        if meta.get('last_modified'):
             image_mtime = dateutil.parser.parse(meta['last_modified'])
-            signature_url = '{}/{}'.format(self.base, meta['signature'])
-            if meta.get('extracted_signature'):
-                extracted_sig_url = '{}/{}'.format(self.base, meta['extracted_signature'])
-            else:
-                extracted_sig_url = None
         else:
-            assert(url is not None and size is not None)
-            image_url = url
-            image_size = size
             image_mtime = None
-            signature_url = url + '.asc'
-            if '.xz' in signature_url:
-                extracted_sig_url = signature_url.replace('.xz', '')
-            elif '.gz' in signature_url:
-                extracted_sig_url = signature_url.replace('.gz', '')
-            else:
-                extracted_sig_url = None
+        signature_url = '{}/{}'.format(self.base, meta['signature'])
 
         image_dir = signature_dir = ''
         if self.args.mirror:
@@ -217,7 +202,7 @@ class EosDownloadImage(object):
         self.download(image_url, image, src_size=image_size,
                       src_mtime=image_mtime)
 
-        if meta is not None and meta.get('extracted_size'):
+        if meta.get('extracted_size'):
             with open(image + '.size', 'w') as f:
                 f.write(str(meta['extracted_size']))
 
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-download-image
```diff
@@ -173,12 +177,30 @@ class EosDownloadImage(object):
...
 
-    def download_and_verify(self, meta, what):
+    def download_and_verify(self, meta, what, url=None, size=None):
+        """ Download and verify image. Caller should either use meta or specify url and size."""
         destdir = self.args.outdir
-        image_url = '{}/{}'.format(self.base, meta['file'])
-        image_size = meta['compressed_size']
-        image_mtime = dateutil.parser.parse(meta['last_modified'])
-        signature_url = '{}/{}'.format(self.base, meta['signature'])
+        if meta is not None:
```
wjt: I don't really like this design: I think a `meta` object should be synthesised by the caller rather than adding mutually-exclusive arguments to this function.

eos-tech-support/eos-download-image
```diff
@@ -267,6 +288,10 @@ class EosDownloadImage(object):
                        help='Image version (default: newest)')
         g.add_argument('-I', '--iso', action='store_true',
                        help='Fetch an ISO image, not a raw disk image')
+        g.add_argument('-u', '--url',
```
wjt: I think `--url` should be in the mutually-exclusive group `g` with `--mirror` and `--windows-tool`: this is conceptually a fourth mode, and this option really is mutually-exclusive with those two. (Ideally each mode would be a subparser but that ship sailed long ago!)

eos-tech-support/eos-download-image
```diff
@@ -267,6 +288,10 @@ class EosDownloadImage(object):
...
+                       help='Fetch compressed image by URL (use with --size)')
+        g.add_argument('-s', '--size',
```
wjt: Add either `type=int`, or define this higher up the file and add `type=positive_int`

```python
def positive_int(s):
    value = int(s)
    if value <= 0:
        raise argparse.ArgumentTypeError(s + ' is not positive')
    return value
```

eos-tech-support/eos-download-image
```diff
@@ -327,8 +352,16 @@ class EosDownloadImage(object):
...
         eos-write-live-image.'''
 
-        _, images = self.fetch_manifest()
         args = self.args
+        if args.url is not None:
+            if args.size is None:
+                log("--size must be used when using --url")
+                sys.exit(1)
+            what = 'iso' if args.url.endswith('.iso') else 'full'
+            print(self.download_and_verify(meta=None, what=what, url=args.url, size=args.size))
```
wjt: I think this block should be a separate function, `fetch_from_url(self)` and should call `download_and_verify` with a `meta` object it makes up.

eos-tech-support/eos-download-image
```diff
@@ -173,12 +177,30 @@ class EosDownloadImage(object):
...
+        """ Download and verify image. Caller should either use meta or specify url and size."""
         destdir = self.args.outdir
-        image_url = '{}/{}'.format(self.base, meta['file'])
-        image_size = meta['compressed_size']
-        image_mtime = dateutil.parser.parse(meta['last_modified'])
-        signature_url = '{}/{}'.format(self.base, meta['signature'])
+        if meta is not None:
+            image_url = '{}/{}'.format(self.base, meta['file'])
+            image_size = meta['compressed_size']
+            image_mtime = dateutil.parser.parse(meta['last_modified'])
```
wjt: This would be:

```python
if 'last_modified' in meta:
    image_mtime = dateutil.parser.parse(meta['last_modified'])
else:
    image_mtime = None
```

eos-tech-support/eos-download-image
```diff
@@ -173,12 +177,30 @@ class EosDownloadImage(object):
...
+            else:
+                extracted_sig_url = None
+        else:
+            assert(url is not None and size is not None)
+            image_url = url
+            image_size = size
+            image_mtime = None
+            signature_url = url + '.asc'
+            if '.xz' in signature_url:
+                extracted_sig_url = signature_url.replace('.xz', '')
```
wjt: Since you're not downloading the `boot.zip` and `boot.zip.asc` files, there's no need to download the uncompressed signature. It is only used on live USBs created with `eos-write-live-image`, which requires the `boot.zip`.

mwleeds:
@wjt thanks, all feedback addressed

eos-tech-support/eos-download-image
```diff
@@ -273,6 +284,9 @@ class EosDownloadImage(object):
         if args.product == 'fnde' and args.personality == 'base':
             args.personality = 'fnde_aluno'
 
+        if args.url.startswith(INTERNAL_BASE_URL):
```
wjt: What if `args.url` is `None` because `--url` was not specified?

```console
$ git checkout T20033-add-url-and-size
Branch T20033-add-url-and-size set up to track remote branch T20033-add-url-and-size from origin.
Switched to a new branch 'T20033-add-url-and-size'
  tensionsheet ~/src/endlessm/eos-meta
$ eos-tech-support/eos-download-image 
Traceback (most recent call last):
  File "eos-tech-support/eos-download-image", line 431, in <module>
    main()
  File "eos-tech-support/eos-download-image", line 426, in main
    app = EosDownloadImage(session)
  File "eos-tech-support/eos-download-image", line 287, in __init__
    if args.url.startswith(INTERNAL_BASE_URL):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

I think this check should be something like:

```python
if args.url is not None and args.url.startswith(INTERNAL_BASE_URL):
```

PR #459, opened 2017-12-03
Minor fixes for eos-convert-system{-qa}

mwleeds:


eos-tech-support/eos-convert-system
```diff
@@ -1,5 +1,16 @@
 #!/bin/bash -e
 
+if [ $1 == "-h" ] || [ $1 == "--help" ]; then
+    echo "Usage:"
+    echo "   $0"
+    echo ""
+    echo "Convert an EOS system from using OSTree to using apt for package"
+    echo "management, and make other changes that are generally desired for"
+    echo "development systems, such as making the journal persistent. You"
+    echo "should run this as root."
```
wjt: I think this message is a bit misleading because if a non-Endless employee read this, they might think “great! this will let me use Endless OS like I used to use Ubuntu!”, but in fact these changes are _not_ desired for non-Endless employees. I'd prefer it to mention: (a) if you're not an Endless developer, this will prevent you installing OS updates, and will not really allow you to use `apt`; (b) this change is irreversible; (c) maybe show this message, and request confirmation, when the script is run without `--help`?

mwleeds: good idea

mwleeds:
Addressed @wjt's comment

PR #460, opened 2017-12-08
New eos-enable-coredumps command, reused it from eos-convert-system scripts

mariospr:
The new command fixes a problem in the current way coredumps were meant to
be enabled, which did not work because (1) eos-metrics-instrumentation would
override the kernel.core_pattern setting and (2) coredumps would be still
disabled due to the default 0-byte size limit set in limits.conf.

Besides, having a separate command makes it more convenient to enable
coredumps in non-converted systems.

https://phabricator.endlessm.com/T20331

### DIFF
--- a/eos-tech-support/eos-enable-coredumps
+++ b/eos-tech-support/eos-enable-coredumps
@@ -1,0 +1,50 @@
+#!/bin/bash -e
+#
+# Enables coredumps and adjusts storage path to work with coredumpcl
+#
+# Copyright (C) 2017 Endless Mobile, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+SYSCONFDIR=/etc
+
+if [ $# -gt 1 ] || [ $1 == "-h" ] || [ $1 == "--help" ]; then
+    echo "Usage:"
+    echo "   `basename $0` [sysconfdir]"
+    echo "Where:"
+    echo "   sysconfdir = path to the system configuration (default: /etc)"
+    exit
+elif [ $# -gt 0 ]; then
+     SYSCONFDIR="${1}"
+     shift
+fi
+
+CONF_FILENAME="99-coredump.conf"
+SYSCTL_CONF_FILE="${SYSCONFDIR}/sysctl.d/${CONF_FILENAME}"
+LIMITS_CONF_FILE="${SYSCONFDIR}/security/limits.d/${CONF_FILENAME}"
+
+# Enable coredumps
+mkdir -p $(dirname "${LIMITS_CONF_FILE}")
+cat <<EOF > "${LIMITS_CONF_FILE}"
+# Enable coredumps for regular root (max 256MB) users and root (unlimited)
+*        soft        core        unlimited
+root     hard        core        256000
+EOF
+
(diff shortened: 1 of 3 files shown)


eos-tech-support/eos-enable-coredumps
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -e
```
pwithnall: I’d ideally like to see this written in Python (because yeuch, bash), but this is sufficiently short. (Don’t rewrite it in Python unless you’re a fanatic.)

It’s missing a copyright header.

eos-tech-support/eos-enable-coredumps
```diff
@@ -0,0 +1,31 @@
...
+
+if [ $# -lt 1 ] || [ $1 == "-h" ] || [ $1 == "--help" ]; then
+    echo "Usage:"
+    echo "   `basename $0` sysconfdir"
+    echo "Where:"
+    echo "   sysconfdir = path to the system configuration (e.g. /etc)"
+    exit
+fi
+
+SYSCONFDIR=${1}
```
pwithnall: This could default to `/etc` if unspecified.

eos-tech-support/eos-enable-coredumps
```diff
@@ -0,0 +1,31 @@
...
+
+CONF_FILENAME=90-coredump.conf
```
pwithnall: The pedant in me is saying that instead of hard-coding `90` we should have this script work out the first unused number, and use that, but I think that would over-complicate things. (Don’t make it do that unless you’re a fanatic. If you do make it do that, rewrite in Python because you’re a fanatic, and because doing that in bash would be horrific.)

eos-tech-support/eos-enable-coredumps
```diff
@@ -0,0 +1,31 @@
...
+SYSCTL_CONF_FILE=${SYSCONFDIR}/sysctl.d/${CONF_FILENAME}
+LIMITS_CONF_FILE=${SYSCONFDIR}/security/limits.d/${CONF_FILENAME}
```
pwithnall: Nitpick: Technically these all need quoting in case `${SYSCONFDIR}` has spaces in, but if it does we’re all doomed anyway.

mariospr:
> Needs a copyright header and a default value of /etc for the parameter. The rest of my comments are just noise.

Nearly no scripts in that directory have a header, that's why I didn't add it, but I'll do anyway, I also consider it a good idea regardless.

eos-tech-support/eos-enable-coredumps
mariospr: I'm not a fanatic and I also like python, but I think this still qualifies as "simple enough for bash" :)

Will add the header

eos-tech-support/eos-enable-coredumps
mariospr: Ok

eos-tech-support/eos-enable-coredumps
mariospr: I'm not a fanatic, and not swimming in time at the moment to rewrite this really :-)

Besides, the number was hardcoded already and using things like `50-`, `90-` is common practice anyway so I think it's good enough as an improvement over what we had. Beside, I'm not convinced you really want to work out that number automatically either in all cases.

eos-tech-support/eos-enable-coredumps
mariospr: Exactly. I can add quotes and I actually thought of it, but checking other scripts seemed to me that we normally don't when we refer to system paths like these ones.

However, the paranoid parrot in me would feel more peaceful if quotes would be around, so I think I will add them after all and blame you if someone complains later on.

eos-tech-support/eos-enable-coredumps
```diff
@@ -0,0 +1,50 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+SYSCONFDIR=/etc2
```
pwithnall: `/etc2`?

mariospr: ARGH!

PR #468, opened 2018-01-11
core: remove btrfs-tools

wjt:
Endless OS does not use btrfs.

https://phabricator.endlessm.com/T20709


dbnicholson:
Wow, this goes all the way back to the initial metapackage work done by Collabora.

PR #469, opened 2018-01-12
os-depends: Rename grub-efi-amd64-image to grub-efi-ia32-image

carlocaione:

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -33,7 +33,8 @@ gnome-shell (>= 3.22.3)
 gnupg
 # Could just be i386 amd64, but theoretically supports other arches
 grub2 [!armhf]
-grub-efi-amd64-image-signed [amd64] | grub-efi-amd64-image [i386]
+grub-efi-amd64-image-signed [amd64 i386]
+grub-efi-ia32-image [amd64 i386]
 intel-microcode [!armhf]
 iproute2
 kmod


os-depends
```diff
@@ -33,7 +33,7 @@ gnome-shell (>= 3.22.3)
 gnupg
 # Could just be i386 amd64, but theoretically supports other arches
 grub2 [!armhf]
-grub-efi-amd64-image-signed [amd64] | grub-efi-amd64-image [i386]
+grub-efi-amd64-image-signed [amd64] | grub-efi-ia32-image [i386]
```
wjt: I don't think this is right: we want to have both `grub-efi-amd64-image-signed` (for the EFI-64 bootloader) and `grub-efi-ia32-image` (for the EFI-32 bootloader) in the image, no?

carlocaione: ok, I'm a bit confused then. My reasoning here is that in the `amd64` image (i.e.) we only need `grubx64.efi`. For the `i386` images only `bootia32.efi`. Why do we want to have both in general?

dbnicholson: I'm not fully educated in EFI, but I think @carlocaione is right. On a 64 bit system (amd64), you would only ever need `grubx64.efi`, right? Whereas on a 32 bit system (i386), you would only need `bootia32.efi`. But, of course, these programs don't run on the OS, so maybe I'm wrong.

That said, I think `bootia32.efi` was moved to a separate package entirely so that it wasn't shipped on amd64. If `grub-efi-ia32-image` is installed on amd64, then you'd still have it, which would defeat the purpose.

dbnicholson: I think the argument you could make is that you might want both 32 and 64 bit EFI loaders on i386 because you might be running a 32 bit OS on a 64 bit system. However, you couldn't say the opposite - 64 bit OS on a 32 bit system. So, I think the most pedantically correct would likely be:
```
grub-efi-amd64-image-signed [amd64 i386]
grub-efi-ia32-image [i386]
```

carlocaione: this makes sense. I'll respin the PR.

wjt: No, we explicitly want the 32-bit EFI loader in the 64-bit OS image. This is because there are machines that have a 64-bit CPU but 32-bit EFI firmware.

wjt: https://phabricator.endlessm.com/T13970

carlocaione: jesus, why such a monster does even exist. Alright, then, it looks like we must ship both the packages on both the architectures.

os-depends
```diff
@@ -33,7 +33,8 @@ gnome-shell (>= 3.22.3)
 gnupg
 # Could just be i386 amd64, but theoretically supports other arches
 grub2 [!armhf]
-grub-efi-amd64-image-signed [amd64] | grub-efi-amd64-image [i386]
+grub-efi-amd64-image-signed [!armhf]
+grub-efi-ia32-image [!armhf]
```
dbnicholson: Personally I'd use `[amd64 i386]` as a whitelist in case we ever add another architecture (e.g, `aarch64`). I see `intel-microcode` below this is doing the same thing.

carlocaione: Sounds good. I'll fix `intel-microcode` in a different PR

dsd:
> I still don't totally understand when you'd get EFI 32 bit on a 64 bit system

I'm also not clear on this package rename (we shipped a 32-bit grub binary in a package named amd64? where does the 64-bit grub come from then?). It would have been good to include an explanation or a pointer in the commit message. Can you at least point me in the right direction now?

dbnicholson:
@dsd See https://github.com/endlessm/grub/pull/69. I think that made more sense before @wjt said we still needed to ship the EFI-IA32 version, too. But, again, my EFI knowledge is sketchy.

carlocaione:
> we shipped a 32-bit grub binary in a package named amd64? 

Yes, while reworking GRUB to fix the messy package generation and moving all the EFI images in a place accessible by the (future) upgrade script I also created a new standalone package for the 32-bit grub binary. It wasn't really a needed change, I just felt it was nice-to-fix (tm).

> where does the 64-bit grub come from then?

I agree with @dbnicholson that probably this made more sense before having to ship it also in the `amd64` images. The problem is that as pointed out by @wjt in https://github.com/endlessm/eos-meta/pull/469#discussion_r161769671 some 64-bit CPU platform using a 64-bit OS image still need the `ia32` version of the GRUB EFI image since they ship a 32-bit UEFI, that's why the latest change.

Sorry for the lacking commit message but the initial idea was indeed shipping the `amd64` version in the `amd64` image and the `ia32` image in the `i386` image but then @wjt ruined the party ;)

wjt:
@dbnicholson “I still don't totally understand when you'd get EFI 32 bit on a 64 bit system” – the short answer is “cost”. Helpfully Intel have deleted [this blog post from 2015](http://blogs.intel.com/evangelists/2015/07/22/why-cheap-systems-run-32-bit-uefi-on-x64-systems/) which explained why…

dbnicholson:
@wjt Makes sense. Found it at https://web.archive.org/web/20150804192705/http://blogs.intel.com/evangelists/2015/07/22/why-cheap-systems-run-32-bit-uefi-on-x64-systems/.

PR #470, opened 2018-01-23
os-depends: Add shim to amd64 images

carlocaione:


wjt:
This is a subset of the changes on #472

PR #472, opened 2018-01-23
T14870

carlocaione:

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -68,6 +68,7 @@ python-gi
 python3-gi
 pulseaudio
 pulseaudio-module-bluetooth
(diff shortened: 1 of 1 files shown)


os-depends
```diff
@@ -63,6 +63,7 @@ openssh-client
 ostree
 p7zip-full
 parted
+pesign [amd64]
```
wjt: I guess this tool is small, and I understand the desire to be able to check at the point the bootloader is updated that it's correctly signed, but – I think this check belongs in the ostree (and image) builder. We should never get as far as publishing unsigned bootloaders to end users.

PR #474, opened 2018-02-05
core: remove btrfs-tools

wjt:
Endless OS does not use btrfs.

This is a re-application of ef61b8b which was inadventently reverted by
15b5eb8.

https://phabricator.endlessm.com/T20709


wjt:
b667bdcd674959b20270139733e07f877f916b37 addresses this.

PR #475, opened 2018-02-07
eos-enable-coredumps: Set default `ulimit -c` value for systemd units

pwithnall:
Even if systemd-coredump is enabled, the default coredump size limit for
all systemd units is not infinite, and this prevents things like
eos-updater successfully dumping core, with the message:

   systemd-coredump[1664]: Resource limits disable core dumping for
   process 1573 (eos-updater).

This is frustrating.

Fix it by having eos-enable-coredumps add a systemd configuration file
which ups the default `ulimit -c` value for units to infinity.

See:
https://www.freedesktop.org/software/systemd/man/systemd-system.conf.html
and
https://lists.freedesktop.org/archives/systemd-devel/2013-February/009174.html.


pwithnall:
@mariospr Would you mind reviewing please?

PR #476, opened 2018-02-08
defaults: Associate GNOME Software with ostree repos

joaquimrocha:
This patch sets GNOME Software as the default application for opening
ostree repos in removable drives. It should result in GNOME Software
opening by default when when a USB containing an ostree repo is
inserted.

https://phabricator.endlessm.com/T16950


cosimoc:
LGTM

PR #479, opened 2018-02-21
Improve error messages from eos-write-{installer,image}

wjt:
https://phabricator.endlessm.com/T11946

### DIFF
--- a/eos-tech-support/eos-write-image
+++ b/eos-tech-support/eos-write-image
@@ -36,7 +36,7 @@ EOF
 FORCE=
 REMOVABLE=
 DD_IN_OPTS=(iflag=nonblock)
-DD_OUT_OPTS=(oflag=nonblock conv=fsync)
+DD_OUT_OPTS=(conv=fsync)
 while true; do
     case "$1" in
         -f|--force)
--- a/eos-tech-support/eos-write-image
+++ b/eos-tech-support/eos-write-image
@@ -191,8 +191,13 @@ else
     fi
 
     if [ ! "$REMOVABLE" ]; then
-        REMOVABLE=$(lsblk --raw --nodeps --noheading --output HOTPLUG "$DEVICE")
-        if [ "$REMOVABLE" = 1 ]; then
+        if ! REMOVABLE=$(lsblk --raw --nodeps --noheading --output HOTPLUG "$DEVICE"); then
+            # This can happen if the disk has been "ejected" rather than
+            # unmounting. Writing to it will fail in this case.
+            echo "Unable to detect whether $DEVICE is removable;" \
+                 "writing to it will probably fail." >&2
+            REMOVABLE=1
+        elif [ "$REMOVABLE" = 1 ]; then
             echo "Guessed that $DEVICE is a removable disk;" \
                  "pass --fixed if this is wrong." >&2
         else
--- a/eos-tech-support/eos-write-installer
+++ b/eos-tech-support/eos-write-installer
@@ -25,7 +25,7 @@ ARGS=$(getopt -o i:o:lh -l "installer:,os-image:,latest,help" -n "$0" -- "$@")
 eval set -- "$ARGS"
 
 usage() {
-    cat <<EOF
+    cat >&2 <<EOF
 Usage:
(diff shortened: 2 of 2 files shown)


eos-tech-support/eos-write-image
```diff
@@ -36,7 +36,7 @@ EOF
 FORCE=
 REMOVABLE=
 DD_IN_OPTS=(iflag=nonblock)
-DD_OUT_OPTS=(oflag=nonblock conv=fsync)
+DD_OUT_OPTS=(conv=fsync)
```
wjt: The `oflag=nonblock` flag has been in here since [the initial commit in this repo](https://github.com/endlessm/eos-meta/commit/7590026a55cdc897541abbaefaf6b05305773c2b) but as described in the commit message it causes misleading errors. I can't find the repo that the script was imported from so I do not know if there is some reason why it needs to be there. The [eos-factory-tools version](https://github.com/endlessm/eos-flash-tools/blob/96d6d26933ffa67029638a5c2936af8f00ecc7fd/eos-usb-reflash/99reflash/shutdown.sh#L39-L51) does not include it.

dbnicholson: It was introduced a long time ago in https://github.com/endlessm/eos-build/commit/62867c6, but there's no justification for it. Maybe @rshuler remembers since he made that change. IIRC, someone may have suggested that using nonblocking I/O could make things go faster. Howerver, I think we don't actually want nonblocking behavior since we're blocking on `dd` to see if it completed. Like you say, doing all the I/O nonblocking would just produce unusual errors.

I think what _could_ actually make things go faster is `oflag=direct`, which opens the file with `O_DIRECT`, but that's kind of orthogonal to this change. I recall reading someone's blog that this can actually make a big difference, but I can't find that now.

rshuler: That was copied from the previous version of the script in eos-shell-build, which we have since deleted.  While we could try to restore that repo and see its commit history, my guess is that came from suggested syntax I found on the web, and at this point I'm sure you know the syntax issues better than I, so please proceed with what you feel is best.

eos-tech-support/eos-write-installer
```diff
@@ -27,23 +27,27 @@ eval set -- "$ARGS"
...
-missing, the latest image will be downloaded. If --latest is not specified,
-both --installer and --os-image must be specified.
+missing, the latest image will be downloaded.
 EOF
 }
 
 function check_exists() {
     if [ ! -f "$1" ]; then
-        echo "$1 does not exist or is not a file device"
+        echo "Error: $1 does not exist or is not a file"
```
dbnicholson: IMO, error messages should be redirected to `stderr` like `>&2`. I know you didn't write these messages originally, but it would be nice if you could fix them up.

eos-tech-support/eos-write-image
```diff
@@ -191,8 +191,13 @@ else
     fi
 
     if [ ! "$REMOVABLE" ]; then
-        REMOVABLE=$(lsblk --raw --nodeps --noheading --output HOTPLUG "$DEVICE")
-        if [ "$REMOVABLE" = 1 ]; then
+        if ! REMOVABLE=$(lsblk --raw --nodeps --noheading --output HOTPLUG "$DEVICE"); then
```
dbnicholson: I had to test that setting the variable in an `if` expression worked correctly, but it seems it does. Reading bash(1) again:
>        if list; then list; [ elif list; then list; ] ... [ else list; ] fi
>              The  if  list is executed.  If its exit status is zero, the then
>              list is executed.  Otherwise, each  elif  list  is  executed  in
>              turn,  and  if  its  exit status is zero, the corresponding then
>              list is executed and the command completes.  Otherwise, the else
>              list  is executed, if present.  The exit status is the exit sta‐
>              tus of the last command executed, or zero if no condition tested
>              true.

That does look like it's appropriate since the list is executed.

eos-tech-support/eos-write-installer
wjt: I've pushed a commit that redirects basically everything to `>&2`, which matches (eg) `eos-write-image` too.

PR #484, opened 2018-03-14
os-depends: Add X fbdev driver

starnight:
Found xserver-xorg-video-fbdev - X.Org X server -- fbdev display driver is
missed in the master image.  The driver is needed for efifb.
If this is missed, the X will be black screen with the nouveau driver which goes
with parameters noaccel=1 modeset=0

Therfore, we need to add fbdev back.

https://phabricator.endlessm.com/T21555

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -99,6 +99,7 @@ xserver-xorg
 xserver-xorg-input-libinput
 # Some of these are really specific to x86, but just keep them off of arm for now
 xserver-xorg-video-amdgpu [!armhf]
+xserver-xorg-video-fbdev [!armhf]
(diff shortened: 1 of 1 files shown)


os-depends
```diff
@@ -106,3 +106,4 @@ xserver-xorg-video-qxl [!armhf]
 xserver-xorg-video-radeon [!armhf]
 xserver-xorg-video-vesa [!armhf]
 xserver-xorg-video-vmware [!armhf]
+xserver-xorg-video-fbdev [!armhf]
```
dsd: please keep this list in alphabetical order

PR #487, opened 2018-03-27
Stress mem tools

treitter:
This adds a few tools for measuring aggregate system memory metrics, PSS and RSS memory usage of processes on the system (aggregated by command), and producing stress on a system in ways we might expect users to stress it.

These are designed to produce output suitable for plugging into specially-formatted spreadsheets for analysis and archiving.

NOTE: this commit adds some test files (one of which is nearly 10 MiB). It's generally not good practice to include large data files in git, so please let me know if you'd like to separate them out into a different location. I would definitely recommend against continuing to add files in this git repo and if we avoid it now we can prevent a "git clone" of this repo from always being significantly larger than it would be otherwise.

### DIFF
--- a/eos-tech-support/eos-procmem-sample
+++ b/eos-tech-support/eos-procmem-sample
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+declare -A rss_sums
+declare -A pss_sums
+
+if [ $EUID != 0 ]; then
+    echo "permission denied: must be run as root" >&2
+    exit 1
+fi
+
+for pid in $(ps -ef | awk '{print $2}'); do
+    if [ -f /proc/$pid/smaps ]; then
+        cmd=$(cat /proc/$pid/comm)
+        rss=$(awk 'BEGIN {i=0} /^Rss/ {i = i + $2} END {print i}' \
+                /proc/$pid/smaps)
+        pss=$(awk 'BEGIN {i=0} /^Pss/ {i = i + $2} END {print i}' \
+                /proc/$pid/smaps)
+
+        rss_sums[$cmd]=$((${rss_sums[$cmd]} + $rss))
+        pss_sums[$cmd]=$((${pss_sums[$cmd]} + $pss))
+    fi
+done
+
+for k in ${!rss_sums[@]}; do
+    output=$(echo "$output\n$k\t${pss_sums[$k]}\t${rss_sums[$k]}")
+done
+
+# PSS is the "proportional" set size. In this measure, each shared page size is
+# divided by the number of processes sharing that page, to avoid the
+# double-counting you get with RSS. Thus, PSS values are a fairer measure of
+# memory usage than RSS for processes that heavily share their memory (like
+# Xorg)
+
+echo -e "process name\tPSS (KiB)\tRSS (KiB)"
+# print results sorted by PSS
+echo -e "$output" | sort -rnk 2
--- a/eos-tech-support/eos-sysmem-sampler
+++ b/eos-tech-support/eos-sysmem-sampler
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+echo -en "time\tphys total\tphys free\tphys used\tphys buff/cache\t"
+echo -en "phys avail\tswap total\tswap free\tswap used"
+echo
+
+while : ; do
+    # populate an array with the system memory summary lines from `free`
+    mapfile -t lines < <(free | egrep '^(Mem|Swap)')
+
+    time=$(date +%s)
+
+    phys_total=$(echo ${lines[0]} | cut -d ' ' -f 2)
+    phys_used=$(echo ${lines[0]} | cut -d ' ' -f 3)
+    phys_free=$(echo ${lines[0]} | cut -d ' ' -f 4)
+    phys_buff=$(echo ${lines[0]} | cut -d ' ' -f 6)
+    phys_avail=$(echo ${lines[0]} | cut -d ' ' -f 7)
+
+    swap_total=$(echo ${lines[1]} | cut -d ' ' -f 2)
+    swap_used=$(echo ${lines[1]} | cut -d ' ' -f 3)
+    swap_free=$(echo ${lines[1]} | cut -d ' ' -f 4)
+
+    echo -en "${time}\t${phys_total}\t${phys_free}\t${phys_used}\t"
+    echo -en "${phys_buff}\t${phys_avail}\t${swap_total}\t"
+    echo -en "${swap_free}\t${swap_used}"
+    echo
+
+    sleep 1
+done
--- a/eos-tech-support/eos-user-stress-test
+++ b/eos-tech-support/eos-user-stress-test
@@ -1,0 +1,130 @@
+#!/bin/bash
+
+set -euo pipefail
+
+DATA_FILES_URL=https://github.com/endlessm/eos-qa-test-files/archive/master.zip
+DATA_FILES_BASEDIR=~/eos-qa-test-files-master
+DATA_FILES_ZIP=${DATA_FILES_BASEDIR}.zip
+DATA_FILES=${DATA_FILES_BASEDIR}/user-tests
+SLEEP_APP_LAUNCH=30
+SLEEP_URL_LAUNCH=10
+
+command=""
+if [ $# -gt 0 ]; then
+        command=$1
+fi
+
+check_dep_flatpak() {
+    # relies on set -e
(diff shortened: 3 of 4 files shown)


eos-tech-support/eos-sysmem-sampler
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+echo -en "time\tphys total\tphys free\tphys used\tphys buff/cache\t"
+echo -en "phys avail\tswap total\tswap free\tswap used"
+echo
+
+while : ; do
+	# populate an array with the system memory summary lines from `top`
+	mapfile -t lines < <(top -b -n 1 | egrep '(Mem|Swap)')
```
rshuler: Would this perhaps be more straightforward with `free` rather than `top`?

eos-tech-support/eos-sysmem-sampler
```diff
@@ -0,0 +1,33 @@
...
+	phys_buff=$(echo $line2cleaned | cut -d ' ' -f 7)
+
+	swap_total=$(echo $line3cleaned | cut -d ' ' -f 1)
+	swap_free=$(echo $line3cleaned | cut -d ' ' -f 3)
+	swap_used=$(echo $line3cleaned | cut -d ' ' -f 5)
+	phys_avail=$(echo $line3cleaned | cut -d ' ' -f 7)
+
+	echo -en "${time}\t${phys_total}\t${phys_free}\t${phys_used}\t"
+	echo -en "${phys_buff}\t${phys_avail}\t${swap_total}\t"
+	echo -en  "${swap_free}\t${swap_used}"
```
rshuler: Nit: extra space after "-en".

eos-tech-support/eos-procmem-sample
```diff
@@ -0,0 +1,37 @@
...
+	exit 1
+fi
+
+for pid in $(ps -ef | awk '{print $2}'); do
+    if [ -f /proc/$pid/smaps ]; then
+	cmd=$(cat /proc/$pid/comm)
+        rss=$(awk 'BEGIN {i=0} /^Rss/ {i = i + $2} END {print i}' /proc/$pid/smaps)
+        pss=$(awk 'BEGIN {i=0} /^Pss/ {i = i + $2 + 0.5} END {print i}' /proc/$pid/smaps)
+	# round float numbers since we use integer arithmetic below
+	printf -v rss %.0f $rss
```
rshuler: More tab --> space

eos-tech-support/eos-procmem-sample
```diff
@@ -0,0 +1,37 @@
...
+	printf -v pss %.0f $pss
+
+	rss_sums[$cmd]=$((${rss_sums[$cmd]} + $rss))
+	pss_sums[$cmd]=$((${pss_sums[$cmd]} + $pss))
+    fi
+done
+
+for k in ${!rss_sums[@]}; do
+	output=$(echo "$output\n$k\t${pss_sums[$k]}\t${rss_sums[$k]}")
```
rshuler: More tab --> space

rshuler:
Oh, I did not notice the extra files when I suggested adding this to eos-meta...  We either need to remove or make these files much smaller, or we'll have to exclude from the regular ostree.  I'll comment on the ticket.

eos-tech-support/eos-user-stress-test
```diff
@@ -0,0 +1,113 @@
...
+        check_deps
+        echo "all dependencies fulfilled"
+        ;;
+    light)
+        check_deps_light
+        command_launch flatpak run com.endlessm.animals.en
+        command_launch lowriter --norestore files/chicken.odt
+        url_launch https://www.google.com/search?q=chicken
+        url_launch https://www.youtube.com/watch?v=dfR_LdA3fPI
+        url_launch https://inbox.google.com
```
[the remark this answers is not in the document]
treitter: In my use of the script, I stopped bothering to log in since it's a hassle. I think it makes sense in a realistic test case but I think one that is interactive. This test case is meant to be as automated as possible, so I've replaced this URL with an account-free page.

treitter:
I've addressed all your other concerns as well in this new version. The files have been relocated to the eos-qa-test-files repo and I'm just about to make a pull request for that.

eos-tech-support/eos-procmem-sample
```diff
@@ -0,0 +1,37 @@
...
+if [ $EUID != 0 ]; then
+    echo "permission denied: must be run as root" >&2
+    exit 1
+fi
+
+for pid in $(ps -ef | awk '{print $2}'); do
+    if [ -f /proc/$pid/smaps ]; then
+    cmd=$(cat /proc/$pid/comm)
+        rss=$(awk 'BEGIN {i=0} /^Rss/ {i = i + $2} END {print i}' /proc/$pid/smaps)
+        pss=$(awk 'BEGIN {i=0} /^Pss/ {i = i + $2 + 0.5} END {print i}' /proc/$pid/smaps)
```
rshuler: I'm pretty sure that this `+ 0.5` is incorrect.  It appears to me that the Pss values are integers to begin with, and you are summing an extra 0.5 for each Pss entry in the file, effectively reporting values that are quite a bit higher than they should be.

eos-tech-support/eos-procmem-sample
```diff
@@ -0,0 +1,37 @@
...
+declare -A pss_sums
+
+if [ $EUID != 0 ]; then
+    echo "permission denied: must be run as root" >&2
+    exit 1
+fi
+
+for pid in $(ps -ef | awk '{print $2}'); do
+    if [ -f /proc/$pid/smaps ]; then
+    cmd=$(cat /proc/$pid/comm)
```
rshuler: Fix indentation for this entire `then` paragraph

eos-tech-support/eos-procmem-sample
```diff
@@ -0,0 +1,37 @@
...
+        rss=$(awk 'BEGIN {i=0} /^Rss/ {i = i + $2} END {print i}' /proc/$pid/smaps)
+        pss=$(awk 'BEGIN {i=0} /^Pss/ {i = i + $2 + 0.5} END {print i}' /proc/$pid/smaps)
+    # round float numbers since we use integer arithmetic below
```
rshuler: I'm pretty sure you can remove this rounding, since all numbers should be integer if you remove the `+ 0.5`

eos-tech-support/eos-user-stress-test
```diff
@@ -0,0 +1,122 @@
...
+
+check_dep_flatpak() {
+    # relies on set -e
+    flatpak info $1 > /dev/null
+}
+
+check_deps_common() {
+    if [ ! -e $DATA_FILES ]; then
+        echo >&2 "missing data files at $DATA_FILES"
+        exit 1
```
rshuler: Please try to wget the files and put them in the correct place, and only exit if the files are unavailable (e.g., if no network connection).

eos-tech-support/eos-procmem-sample
treitter: You're right. Fixed in the new version.

eos-tech-support/eos-procmem-sample
treitter: Fixed in the new version.

eos-tech-support/eos-procmem-sample
treitter: Fixed in the new version.

eos-tech-support/eos-user-stress-test
treitter: Fixed in the new version. I'm just downloading the master zip file for the git repo since it seems like a relatively stable URL. In the worst case, ensuring the test files already exist skips the download phase.

eos-tech-support/eos-user-stress-test
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+
+set -euo pipefail
+
+DATA_FILES_URL=https://github.com/endlessm/eos-qa-test-files/archive/master.zip
```
rshuler: This file does not exist in the repo.  Personally, I would avoid using a zip file, as that is just more work for the developer to maintain the files, and easily forgotten to update.  I don't see an easy way to download a directory of files from GitHub, but since the script knows exactly which files it needs, you could just download the four files separately.

rshuler: Ah, as Travis pointed out, the web server creates this zip file.  As long as we are not putting other stuff into this repo, this makes sense for now.  I'll merge as is.

PR #489, opened 2018-03-30
eos-stage-flatpak: avoid failure when --user repos are configured

ramcq:
As this command directly acts on the system ostree repo, if flatpak lists any user remotes, it fails.

### DIFF
--- a/eos-tech-support/eos-stage-flatpak
+++ b/eos-tech-support/eos-stage-flatpak
@@ -57,7 +57,7 @@ else
     flatpak_repo=/var/lib/flatpak/repo
 fi
 
-for remote in `flatpak remote-list | awk {'print $1'}` ; do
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-stage-flatpak
```diff
@@ -57,7 +57,7 @@ else
     flatpak_repo=/var/lib/flatpak/repo
 fi
 
-for remote in `flatpak remote-list | awk {'print $1'}` ; do
+for remote in `flatpak remote-list --system | awk {'print $1'}` ; do
```
rshuler: Sounds good.  The awk might not strictly be needed anymore, but I suppose it doesn't hurt.

rshuler:
Accepted and cherry-picked to eos3.3.

PR #491, opened 2018-04-05
Only include the previous and current boot in journal

treitter:
By default, exclude journal records before the previous and current
boot.

This is to reduce the amount of potentially-private information exposed
by default as we will shortly enable persistent storage by default (for
systems with durable storage devices).

https://phabricator.endlessm.com/T21771

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -691,10 +691,23 @@ let diagnostics = [
         hardwareInfo: true,
         coredumpInfo: true,
         content: function(verboseFlag) {
+            let verboseAppend = ''
             if (verboseFlag)
-                return trySpawn('journalctl --no-pager --output verbose');
-            else
-                return trySpawn('journalctl --no-pager');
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-diagnostics
```diff
@@ -872,6 +886,10 @@ ARGV.forEach(function(arg) {
         coredumpsFlag = true;
     else if (arg == '-v' || arg == '--verbose')
         verboseFlag = true;
+    else if (arg == '-f' || arg == '--full-journal')
+        // only include the full journal if explicitly requested to limit
+        // exposure of potentially-private information by default
```
ramcq: This comment is misleading. The purpose of not including the full journal by default is purely to avoid producing gigantic files. There is not appreciably less privacy risk from including just two boots.

ramcq: (The commit message is also similarly confused!)

treitter:
@ramcq I force-pushed an update which addresses the comment and commit message

PR #497, opened 2018-04-18
T22113: tidy up spelling packages

ramcq:
Tidying a few loose ends from https://phabricator.endlessm.com/T22113 - due to the order of apt processing eos-core dependencies, aspell-en was being installed un-necessarily, myspell is deprecated in favour of hunspell, and we started shipping myspell-pt by mistake.

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -160,7 +160,6 @@ libpam-gnome-keyring
 libsasl2-modules
 modemmanager
 myspell-es
-myspell-pt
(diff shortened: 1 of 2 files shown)


ramcq:
Turns out this aspell vs hunspell has been filed upstream since a year ago, and it looks like enchant is just not that actively maintained. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=860888

PR #503, opened 2018-04-25
eos-fix-ostree-repo: Use the correct enum name

GeorgesStavracas:
Our OSTree bindings currently expose OSTree.RepoCommitState
enums as NORMAL and PARTIAL, while we were using the old
value REPO_COMMIT_STATE_PARTIAL.

Fix that by using PARTIAL instead.

### DIFF
--- a/eos-tech-support/eos-fix-ostree-repo
+++ b/eos-tech-support/eos-fix-ostree-repo
@@ -249,7 +249,7 @@ def mark_commits_partial(repo):
         if objtype != OSTree.ObjectType.COMMIT:
             continue
         _, commit, state = repo.load_commit(checksum)
-        if state == OSTree.RepoCommitState.REPO_COMMIT_STATE_PARTIAL:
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-fix-ostree-repo
```diff
@@ -249,7 +249,7 @@ def mark_commits_partial(repo):
         if objtype != OSTree.ObjectType.COMMIT:
             continue
         _, commit, state = repo.load_commit(checksum)
-        if state == OSTree.RepoCommitState.REPO_COMMIT_STATE_PARTIAL:
+        if state == OSTree.RepoCommitState.PARTIAL:
```
dbnicholson: I think we'll need to add compatibility for the old names since we distribute this script to people on old Endless installs. See all the other junk earlier in the script trying to cope with missing symbols.

pwithnall: ⇒ #505

PR #505, opened 2018-04-25
eos-fix-ostree-repo: Make it work on older OSTrees

GeorgesStavracas:
Users with older OSTree versions might be using this
script, so make sure OSTree.RepoCommitState.PARTIAL is
always defined.


eos-tech-support/eos-fix-ostree-repo
```diff
@@ -67,6 +67,18 @@ import sys
...
+# Older OSTree versions had the GI annotation wrong, and the enum
+# value was OSTree.RepoCommitState.REPO_COMMIT_STATE_PARTIAL while
+# recent versions have OSTree.RepoCommitState.PARTIAL.
+#
+# So we try to load OSTree.RepoCommitState.PARTIAL and, if that fails,
+# override OSTree.RepoCommitState.PARTIAL to be the old enum value.
+try:
+    _value = OSTree.RepoCommitState.PARTIAL
+except:
+    OSTree.RepoCommitState.PARTIAL = OSTree.RepoCommitState.REPO_COMMIT_STATE_PARTIAL
```
dbnicholson: I think you can use `dir()` to inspect the attributes of `OSTree.RepoCommitState` without doing an evaluation:
```
if 'PARTIAL' not in dir(OSTree.RepoCommitState):
    OSTree.RepoCommitState.PARTIAL = OSTree.RepoCommitState.REPO_COMMIT_STATE_PARTIAL
```
That worked in my interactive session with a pre-upgrade ostree.

dbnicholson: And for the record, this change was because of https://github.com/ostreedev/ostree/pull/1335.

PR #507, opened 2018-04-30
T22314 and T20929

mwleeds:


eos-tech-support/eos-stage-flatpak
```diff
@@ -69,6 +69,23 @@ for remote in `flatpak remote-list --system | awk {'print $1'}` ; do
         echo "Setting flatpak $remote URL to $new_url"
         flatpak remote-modify "$remote" --url="$new_url"
 
+        # Update the collection ID if it's set
+        collection_id=$(ostree --repo="$flatpak_repo" config get 'remote "'"$remote"'".collection-id' 2>/dev/null || echo "")
+        if [ -n "$collection_id" ] ; then
+            if [ "$STAGE" == "prod" ] ; then
+                new_collection_id=`echo "$collection_id" | sed -e "s/Dev.//" | sed -e "s/Demo.//"`
+            elif [ "$STAGE" == "dev" ] ; then
+                new_collection_id=`echo "$collection_id" | sed -e "s/endlessm./endlessm.Dev./"`
```
rshuler: This does not quite work for the swtich from demo to dev and vice-versa.  E.g.:
```
$ echo com.endlessm.Demo.flatpak | sed -e "s/endlessm./endlessm.Dev./"
com.endlessm.Dev.Demo.flatpak
```

eos-tech-support/eos-stage-flatpak
```diff
@@ -69,6 +69,23 @@ for remote in `flatpak remote-list --system | awk {'print $1'}` ; do
         echo "Setting flatpak $remote URL to $new_url"
         flatpak remote-modify "$remote" --url="$new_url"
 
+        # Update the collection ID if it's set
+        collection_id=$(ostree --repo="$flatpak_repo" config get 'remote "'"$remote"'".collection-id' 2>/dev/null || echo "")
+        if [ -n "$collection_id" ] ; then
+            if [ "$STAGE" == "prod" ] ; then
+                new_collection_id=`echo "$collection_id" | sed -e "s/Dev.//" | sed -e "s/Demo.//"`
```
rshuler: Is it guaranteed that there is something after the ".Dev"?  If not, this won't quite work.  For example...
```
roddy@endless:~/checkout/glib$ echo com.endlessm.Dev | sed "s/Dev.//"
com.endlessm.Dev
```

eos-tech-support/eos-stage-flatpak
```diff
@@ -68,5 +68,20 @@ for remote in `flatpak remote-list --system | awk {'print $1'}` ; do
...
+        for arch in `flatpak --supported-arches`; do
+            ostree --repo="$flatpak_repo" refs --delete "$remote:appstream/$arch"
+            # This might fail for the secondary architecture
+            flatpak update --appstream --arch="$arch" "$remote" >/dev/null 2>&1 || true
+        done
+
+        # Do the same with ostree-metadata
+        # https://github.com/flatpak/flatpak/issues/1477#issuecomment-372167938
+        ostree --repo="$flatpak_repo" refs --delete "$remote:ostree-metadata"
+        flatpak remote-ls "$remote" >/dev/null
```
rshuler: What is the purpose of calling `flatpak remote-ls` here?

eos-tech-support/eos-stage-flatpak
mwleeds: Yes, it's guaranteed

eos-tech-support/eos-stage-flatpak
mwleeds: getting the new ostree-metadata. We could take it out and just let the next operation that needs it get it, and we could do the same with the appstream.

eos-tech-support/eos-stage-flatpak
mwleeds: Oops, good catch

eos-tech-support/eos-stage-flatpak
mwleeds: I'm not sure how GNOME Software reacts to the appstream data changing while it's running, but better to have the right data than none I think.

eos-tech-support/eos-stage-flatpak
pwithnall: Could do with some examples of the formats of collection ID which this will deal with, in the comment above the `sed` line.

eos-tech-support/eos-stage-flatpak
rshuler: I'm not familiar with `flatpak remote-ls` updating the appstream data.  Don't you need `flatpak update --appstream` for that?

mwleeds: Yes, that's why `flatpak update --appstream` is a few lines above that

eos-tech-support/eos-stage-flatpak
```diff
@@ -72,12 +72,17 @@ for remote in `flatpak remote-list --system | awk {'print $1'}` ; do
...
+                # change e.g. com.endlessm.Dev.Apps → com.endlessm.Apps or com.endlessm.Demo.Apps → com.endlessm.Apps
                 new_collection_id=`echo "$collection_id" | sed -e "s/Dev.//" | sed -e "s/Demo.//"`
             elif [ "$STAGE" == "dev" ] ; then
-                new_collection_id=`echo "$collection_id" | sed -e "s/Dev.//" | sed -e "s/Demo.//" | sed -e "s/endlessm./endlessm.Dev./"`
+                # change e.g. com.endlessm.Demo.Apps → com.endlessm.Dev.Apps or com.endlessm.Apps → com.endlessm.Dev.Apps
+                new_collection_id=`echo "$collection_id" | sed -e "s/Dev.//" | sed -e "s/Demo.//" | sed -e "s/com\.endlessm\./com.endlessm.Dev./"`
             elif [ "$STAGE" == "demo" ] ; then
-                new_collection_id=`echo "$collection_id" | sed -e "s/Dev.//" | sed -e "s/Demo.//" | sed -e "s/endlessm./endlessm.Demo./"`
+                # change e.g. com.endlessm.Dev.Apps → com.endlessm.Demo.Apps or com.endlessm.Apps → com.endlessm.Demo.Apps
+                new_collection_id=`echo "$collection_id" | sed -e "s/Dev.//" | sed -e "s/Demo.//" | sed -e "s/com\.endlessm\./com.endlessm.Demo./"`
```
pwithnall: If escaping the `.`s` is necessary for `sed` (I can never remember and haven’t checked), then it needs to be done for all the other `.`s and `sed` invocations too.

mwleeds: Yeah, here's proof that it's necessary:
```
$ echo "comMendlessm.Apps" | sed -e "s/com.endlessm/com.endlessm.Dev/"
com.endlessm.Dev.Apps
```

PR #508, opened 2018-05-02
Update eos-dev dependencies

dsd:
Add an explanatory comment in eos-dev-depends to explain what types of
packages should be listed there.

Remove the following packages which I assume do not meet these guidelines:
bison, bonnie++, elinks, flex, fluxbox, germinate, glade, gparted,
gperf, language-pack-*, lftp, libjson-glib-doc, locate, ltrace, moreutils,
nethogs, nmap, putty, sl, tig

gnome-api-docs is not present in the distro.

devhelp is available as a flatpak.

openssh-server and p7zip-full are included in the main distro

Switch pyflakes to python3 version

Philip C suggests that eos-node-modules-dev is probably not used
by anyone any more, since you can do node development on unconverted
systems.

Add some packages that I think should be here:
manpages, obs-build, pbuilder

Create and document a new eos-dev-suggests file, to list packages
that do not meet the eos-dev-depends guidelines, but should not be
removed from the distro as they are still valuable developer tools for
limited/occasional use.
These packages will then not be considered for deletion from our distro
if they are not needed by anything else.

### DIFF
--- a/eos-metapackage
+++ b/eos-metapackage
@@ -86,6 +86,7 @@ sub subst_depends {
 foreach my $package (@{$dh{DOPACKAGES}}) {
 	subst_depends($package, "depends", "eos:Depends");
 	subst_depends($package, "recommends", "eos:Recommends");
+	subst_depends($package, "suggests", "eos:Suggests");
 }
 
 =head1 SEE ALSO
--- a/debian/control
+++ b/debian/control
@@ -167,6 +167,7 @@ Package: eos-dev
 Architecture: all
 Depends: ${misc:Depends}, ${eos:Depends}
 Recommends: ${eos:Recommends}
+Suggests: ${eos:Suggests}
 Description: EndlessOS Development Tools
  This is a collection of development tools that make development and debugging
  of EndlessOS easier and more efficient.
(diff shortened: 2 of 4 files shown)


dsd:
@dbnicholson please review this and merge if you agree with the approach.
After merge my plan is to send an endless-dev email highlighting this change in case I made bad assumptions, and giving people a chance to state their interest in the ~400 unused packages that I've identified can be removed from OBS, before I go further in the distro purge.

dbnicholson:
Wow, thank you for doing this. eos-dev was such an eyesore. This was basically one person's idea of "here's a bunch of stuff developers might want".

We should also consider making the eosdev ostree and images again after these changes, but that's a separate topic.

eos-metapackage
```diff
@@ -86,6 +86,7 @@ sub subst_depends {
 foreach my $package (@{$dh{DOPACKAGES}}) {
 	subst_depends($package, "depends", "eos:Depends");
 	subst_depends($package, "recommends", "eos:Recommends");
+	subst_depends($package, "suggests", "eos:Suggests");
```
dbnicholson: Can you put this change in a separate commit? Probably doesn't matter, but it's independent and should stay even if we have to revert the rest of this for some reason.

dsd:
updated now

PR #510, opened 2018-05-08
Add developer script to enable Debian repo

dsd:
Sometimes, a developer tool is not available in OBS but it can be
installed from official debian packages without dependency headaches.

Add a developer-only script that adds apt config for the regular Debian
repos, configured in such a way that Endless stuff takes precedence.

Useful reading:
https://wiki.debian.org/AptPreferences
https://gist.github.com/JPvRiel/8ae81e21ce6397a0502fedddca068507

### DIFF
--- a/eos-tech-support/eos-enable-debian-apt
+++ b/eos-tech-support/eos-enable-debian-apt
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -e
+
+echo "This script adds the Debian repository to your apt config."
+echo "The resulting configuration prefers the Endless repository where"
+echo "possible, even if a package in Endless is older than in Debian."
+echo "However, but packages not found in the Endless repo will be automatically"
+echo "downloaded from Debian".
+echo
+echo "Although this may work fine in some cases, you're likely to find issues"
+echo "where a package in Debian is somehow incompatible with Endless packages"
+echo "and this will most frequently appear as dependency resolution issues."
+echo "As such, this is an unsupported, developer-only tool."
+echo
+echo "Additionally, apt will only work after you have taken the usual steps of"
+echo "making it available (e.g. eos-convert-system, eos-dev-unlock, etc)"
+echo
+
+if [[ $# < 1 ]]; then
+	echo "Usage: $0 <DEB_DISTRO>"
+	echo "DEB_DISTRO example values: stable, testing, unstable"
+	exit 1
+fi
+
+distro=$1
+
+cat > /etc/apt/preferences.d/eos-enable-debian-apt <<EOF
+# Prioritise packages from Debian lower than the default value of 500.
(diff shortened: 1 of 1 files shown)


dsd:
@dbnicholson what do you think?

eos-tech-support/eos-enable-debian-apt
```diff
@@ -0,0 +1,42 @@
...
+	echo "DEB_DISTRO example values: stable, testing, unstable"
+	exit 1
+fi
+
+distro=$1
+
+cat > /etc/apt/preferences.d/eos-enable-debian-apt <<EOF
+Package: *
+Pin: release o=Endless
+Pin-Priority: 600
```
dbnicholson: I don't think you really need to raise this above the default priority of 500 since you're explicitly lowering the debian priority to 400 below. I don't think it hurts anything in normal operation, but it's a little unexpected.

dsd: I agree, this line can be removed

eos-tech-support/eos-enable-debian-apt
```diff
@@ -0,0 +1,42 @@
...
+
+Package: *
+Pin: release o=Debian
+Pin-Priority: 400
+EOF
+
+cat > /etc/apt/sources.list.d/debian.list << EOF
```
ramcq: Maybe debian-$distro.list so if you want to get very funky you can use multiple?

eos-tech-support/eos-enable-debian-apt
ramcq: Yeah I agree, Debian to 400 should be sufficient. Did you verify that dist-upgrade doesn't do anything whacky after adding Debian eg stable? ie the pin seems to do what we expect?

eos-tech-support/eos-enable-debian-apt
```diff
@@ -0,0 +1,42 @@
...
+distro=$1
+
+cat > /etc/apt/preferences.d/eos-enable-debian-apt <<EOF
+Package: *
+Pin: release o=Endless
+Pin-Priority: 600
+
+Package: *
+Pin: release o=Debian
+Pin-Priority: 400
```
dbnicholson: On a related note, it would be nice to add a comment above this block explaining why the 400 was picked (make it below the default 500 so debian packages are only installed if they aren't on Endless or the debian release was specified).

dsd:
updated the request

dsd:
It'll be installed as part of eos-tech-support

dbnicholson:
Oh, right. We just wildcard that whole directory. Not sure how I feel about having this be part of our OS. It kinda implies that we support installing random packages from debian. I'd prefer if this was in a developer focused repo that's not part of the OS.

PR #515, opened 2018-06-07
eos-download-image: attempt to handle truncated responses

wjt:
Mazen's [seeing][1] partial downloads of images, followed by (of course) GPG verification failing. I'm not sure why; my hypothesis is that CloudFront is serving a short response because its connection to S3 drops.  This is supported by [Amazon's documentation][0]:

> **Dropped TCP Connections**
>
> If the TCP connection between CloudFront and your origin drops while your origin is returning an object to CloudFront, CloudFront behavior depends on whether your origin included a Content-Length header in the response:
>
> - Content-Length header – CloudFront returns the object to the viewer as it gets the object from your origin. However, if the value of the Content-Length header doesn't match the size of the object, CloudFront doesn't cache the object.

So add a check that each request returned the full amount of data expected; if it does not, push a new range back onto the queue. Also add a check at the very end that the full length of the file has been fetched.

I also noticed that resuming >20GB files was broken; fix this too.

[0]: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html
[1]: https://phabricator.endlessm.com/P547

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # vim: fileencoding=utf-8 sts=4 sw=4 et
 import argparse
+import collections
 import datetime
 import dateutil.parser
 import distutils.version
@@ -131,7 +132,7 @@ class EosDownloadImage(object):
 
         # Break the download up into ranges if it exceeds the CDN maximum
         # size.
-        ranges = []
+        ranges = collections.deque()
         start_byte = dest_size
         remaining = src_size - dest_size
         while remaining > 0:
@@ -153,7 +154,8 @@ class EosDownloadImage(object):
 
             received = dest_size
             progress(dest, int(100 * received / src_size))
-            for start, stop in ranges:
+            while ranges:
+                start, stop = ranges.popleft()
                 headers['Range'] = 'bytes={}-{}'.format(start, stop)
                 resp = self.session.get(src, headers=headers, stream=True,
                                         timeout=TIMEOUT)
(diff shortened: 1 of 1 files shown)


wjt:
@mazen-asef please try this branch and see if it works better for you!

mazen-asef:
Hey @wjt ! Managed to successfully download an iso :)


Downloading iso image signature from https://d1anzknqnc1kmb.cloudfront.net/nightly/solutions-amd64-amd64/eos3.4/brazil_moe_demo/180606-054213/solutions-eos3.4-amd64-amd64.180606-054213.brazil_moe_demo.iso.asc
solutions-eos3.4-amd64-amd64.180606-054213.brazil_moe_demo.iso.asc (0%)
Range: bytes=0-832
solutions-eos3.4-amd64-amd64.180606-054213.brazil_moe_demo.iso.asc (100%)

Verifying

$ gpgv --keyring /usr/share/keyrings/eos-image-keyring.gpg solutions-eos3.4-amd64-amd64.180606-054213.brazil_moe_demo.iso.asc solutions-eos3.4-amd64-amd64.180606-054213.brazil_moe_demo.iso
gpgv: Signature made qua 06 jun 2018 03:31:24 -03
gpgv:                using RSA key CB500F7BC9233FAD32B4E7209E0C1250587A279C
gpgv: Good signature from "Endless Image Signer 1 (EIS1) <maintainers@endlessm.com>"

Download okay.
solutions-eos3.4-amd64-amd64.180606-054213.brazil_moe_demo.iso

dbnicholson (approved):
Makes perfect sense to me. I hadn't thought of it before, but you should for sure check that the total number of bytes received matches the content length.

PR #522, opened 2018-06-26
eos-download-image: use images-dl.endlessm.com, handle redirects

wjt:
By default, requests.head() does not follow redirects, so the response
when trying to get the file size did not contain Content-Length.
images-dl.endlessm.com is currently a CNAME rather than a redirector,
but we should pre-emptively handle that case.

https://phabricator.endlessm.com/T16529

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -14,7 +14,7 @@ import sys
 import urllib.parse
 
 INTERNAL_BASE_URL = "http://images.endlessm-sf.com"
-PUBLIC_BASE_URL = "https://d1anzknqnc1kmb.cloudfront.net"
+PUBLIC_BASE_URL = "https://images-dl.endlessm.com"
 KEYRING_FILENAME = "eos-image-keyring.gpg"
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-download-image
```diff
@@ -79,8 +79,13 @@ class EosDownloadImage(object):
...
+        resp = self.session.head(src, headers=headers, timeout=TIMEOUT,
+                                 allow_redirects=True)
         if resp.status_code == 200:
+            if src != resp.url:
+                log("Followed redirect to", resp.url)
+                src = resp.url
+
             src_mtime = dateutil.parser.parse(resp.headers['Last-Modified'])
             src_size = int(resp.headers['Content-Length'])
         elif resp.status_code == 403:
```
wjt: It occurs to me that by pointing to a server we control, rather than directly to CloudFront, we could actually fix this problem where HEAD is not supported, by getting the Content-Length from s3fs.

dbnicholson: I'm not sure where to keep track of that idea, but it's a good one!

PR #528, opened 2018-08-06
eos-core: add libu2f-udev

wjt:
This sets permissions on USB U2F tokens so they can be used by unprivileged processes.

I also alphabetized the few unalphabetized entries in this file, as a later patch in case we want to cherry-pick the functional change back to eos3.4.

This is the approach taken by Debian and its descendents. (https://github.com/endlessm/eos-meta/pull/527 is an alternative approach, following Fedora's approach.)

https://phabricator.endlessm.com/T23496

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -37,9 +37,9 @@ debconf-i18n
 dnsmasq-base
 # For brasero video DVD functionality
 dvdauthor
-eog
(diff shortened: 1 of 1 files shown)


wjt:
ok to test

wjt:
We decided against this approach.

PR #529, opened 2018-08-07
eos-tech-support: collect `ostree refs`

wjt:
While this largely duplicates the two `flatpak list` incantations, it
would have allowed us to notice sooner that the local:factory and
eos:eos2/i386 refs were still around on a system upgraded from eos2. We
could try to filter out known flatpak remotes from this list but it
doesn't seem worth it. OTOH it *does* seem worth sorting the list
because it's otherwise illegible.

https://phabricator.endlessm.com/T23443


wjt:
Might be nice to know how much data is uniquely reachable from each ref but I assume that is not easy.

PR #534, opened 2018-10-16
eos-diagnostics: Trim '\0' from DT compatible string

carlocaione:
https://phabricator.endlessm.com/T23912


eos-tech-support/eos-diagnostics
```diff
@@ -53,6 +53,7 @@ function collectBoardInfo() {
     try {
         let [res, contents] = GLib.file_get_contents('/proc/device-tree/compatible');
         output += 'compatible: ' + contents;
+        output = output.replace(/\0/g, ' ');
```
wjt: Well, I wonder if we should do this at the end of the function on the whole of `output`, unconditionally. Is it guaranteed that the DMI files don't contain embedded NULs?

PR #536, opened 2018-11-13
T24374 hack reset

dnohales:
## Don't merge this yet

This is for early review, I still need to investigate the debian packaging part of this.

https://phabricator.endlessm.com/T24374


eos-tech-support/eos-hack-reset-first-contact
```diff
@@ -0,0 +1,42 @@
...
+
+rm $flag_file
+
+IFS=$'\n'
+for user_row in $(cat /etc/passwd); do
+    user=$(echo $user_row | cut -d : -f 1)
+    user_id=$(echo $user_row | cut -d : -f 3)
+
+    if [[ $user_id -ge 1000 && $user_id -le 1999 ]]; then
+        delete_tries=1
```
ramcq: What is the thinking behind the retry loop? Under what circumstances are you expecting userdel to fail, and later succeed?

eos-tech-support/eos-hack-reset-first-contact
```diff
@@ -0,0 +1,42 @@
...
+        while true; do
+            echo "Deleting user $user ($user_id) (try: $delete_tries)"
+            userdel $user
```
ramcq: Is it intentional that you are removing the user but not the actual home directory? This could have very weird/surprising effects - if you recreate user later with the same name, but they have a different UID, they will not own their home directory and then their user account will likely be unusable.

eos-tech-support/eos-hack-reboot-to-first-contact
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -e
+
+usage () {
+    cat <<EOF
+Usage: $0
+Reboot the machine and deletes all users in the system,
+so you can execute the first contact experience from
```
ramcq: I would just name this something like eos-user-reset or eos-erase-users similar, as it's not specific to Hack and could be a useful CS tool generally.

eos-tech-support/eos-hack-reset-quest
```diff
@@ -0,0 +1,60 @@
...
+Reset the state of QUEST-NAME, you can specify "all",
+to do a full reset of all the quests and a reset of the
+desktop icons generated by quests.
+
+  -h, --help                display this help and exit
+
+EOF
+}
+
+args=$(getopt -o h -l "help" -n "eos-hack-reset-quests" -- "$@")
```
ramcq: Nitpick: this -n is inconsistent with the binary name.

ramcq: ($0 will always match!)

wjt: `$(dirname "$0")` is even better!

eos-tech-support/eos-hack-reset-quest
```diff
@@ -0,0 +1,60 @@
...
+    state_filename=~/.var/app/com.endlessm.GameStateService/data/state.json
+
+    flatpak run --command=true com.endlessm.HackToolbox 2> /dev/null
+    if [ $? != 0 ]; then
+        echo "You have to install 'com.endlessm.HackToolbox' in order to reset a single quest"
+        exit 1
+    fi
+
+    flatpak run --command=jq --filesystem=~/.var/app/com.endlessm.GameStateService \
+        com.endlessm.HackToolbox \
```
wjt: This is weird: run app A, giving it access to app B's state, and manually modify the config file. This ties the file format of GameStateService (maintained outside the OS) to this script (maintained in the OS) so you're committing to supporting this file format forever. I think it would be better to ship this tool as part of GameStateService and just have this OS-level script run `flatpak run --command=eos-hack-reset com.endlessm.GameStateService "$@"`.

eos-tech-support/eos-hack-reset-first-contact
wjt: Our FAQ for deleting all users uses `userdel -r` https://support.endlessm.com/hc/en-us/articles/360006827311-How-do-I-delete-all-users-from-my-computer-

eos-tech-support/eos-hack-reboot-to-first-contact
```diff
@@ -0,0 +1,57 @@
...
+done
+
+# Check that script was run with superuser privileges.
+if [[ $EUID != 0 ]]; then
+    echo "$0 requires superuser privileges."
+    exit 1
+fi
+
+echo "This script will reboot the system and DELETE ALL USERS so you can"
+echo "restart the Hack first contact experience and create a new user."
```
wjt: It may be worth noting here that this script will dump you at the Endless logo for a potentially long period of time while the script runs with no user-visible feedback.

eos-tech-support/eos-hack-reset-first-contact
dnohales: When you isolate `eos-hack-reset-first-contact.target`, the processes used by the users take some time to finish.

ramcq: Could you achieve similar with After=systemd-kill-everything? (Whatever it's called.)

eos-tech-support/eos-hack-reset-first-contact
dnohales: I see, well, I wanted to be not that destructive, most of the time the new user, if you are going to use the same name, is gonna match the ID if that was your only user in the system. But you are right, let's clean it properly.

dnohales:
Refactored the user deletion scripts so they are not Hack specific. Now they are named `eos-factory-reset*`. I structured these scripts so is feasible to add more "reset stages" like cleaning `/etc` or deleting/reinstalling apps in the future.

Added more flags so the user can select which stage to reset and a flag to disable user confirmation (so we can add a GUI for factory reset in the future).

Also, I moved the systemd files here: https://github.com/endlessm/eos-boot-helper/pull/247 . And started thinking that the scripts should be moved there as well for simplicity.

ramcq:
I agree both sides of the feature should live in the same repo (ie: over there)

ramcq:
Oops I didn't mean to hit close.

eos-factory-reset/eos-factory-reset
```diff
@@ -0,0 +1,93 @@
...
+            exit 0
+            ;;
+        --)
+            shift
+            ;;
+        *)
+            break
+            ;;
+    esac
+done
```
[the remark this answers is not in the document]
dnohales: I hate shell scripting, I've chosen it because there were already many of them in the repo.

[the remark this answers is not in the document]
wjt: I fully support the slogan “No More Non-Trivial Shell Scripts”

eos-tech-support/eos-hack-reset-quest
dnohales: I just wanted to use `jq` :(

dnohales:
Gonna move this script to `hack-game-state-service`, closing.

PR #538, opened 2018-11-15
Add script to recover corrupted partition table

jprvita:
On certain setups eos-reclaim-swap may change the start of the bootable partition, leaving the system un-bootable.

This PR adds a script which can be used by affected users from a live boot environment to recover from this problem.

https://phabricator.endlessm.com/T24493


eos-tech-support/eos-fix-mbr-swap-removal
```diff
@@ -44,22 +48,43 @@ exit_error() {
...
-
-# Check this is a MBR partiton table
-label_type=$(sed -n -e 's/^label: [ ]*\(.*\)$/\1/p' ${PARTITIONS_FILE})
-if [[ ${label_type} != "dos" ]] ; then
-    exit_error "Error: This is not a MBR partition table." 2
+if [[ ${#} -eq 1 ]] ; then
+    disks=${1}
+    print_log "Using disk specified by user: ${disks}."
+else
+    disks=$(echo $(lsblk --nodeps --noheadings --list --output NAME  | sed -e 's:\(.*\):/dev/\1:'))
```
dsd: $(echo seems redundant, can you just do disks=$(lsblk ....

eos-tech-support/eos-fix-mbr-swap-removal
```diff
@@ -17,11 +17,15 @@
...
 
-if [[ ${#} -ne 1 ]] ; then
-    echo "Usage: ${0} DEVICE"
+if [[ ${#} -gt 1 ]] ; then
+    echo "Usage: ${0} [DEVICE]"
+    echo "If DEVICE is not specified, we will attempt to auto-detect it."
     exit 1
 fi
 
+PARTITION_START=131072
```
dsd: this variable is unused

eos-tech-support/eos-fix-mbr-swap-removal
```diff
@@ -1,4 +1,21 @@
...
+# Licensed under the GPLv2.
+#
+# eos-fix-mbr-swap-removal: Fix a MBR partition table that was corrupted during
+# swap reclaiming.
+#
+# Since we disabled the use of a swap partition on Endless systems, the
+# eos-reclaim-swap service removes the unused swap partition, if existent, and
+# extents the root partition and filesystem to make use of that space. But on
+# systems which were originally installed from an ISO image prior to Endless
+# 3.4.0 that was booted in Legacy (non-UEFI) mode, eos-reclaim-swap ends up
```
dsd: The triggering conditions listed here are incomplete. I'm going to add some more info to T24493.

Perhaps we can just be unspecific here, I think the rest of the comment is giving enough context about the script. "On some old Endless systems installed under certain conditions, eos-reclaim swap ends up corrupting..." or something

eos-tech-support/eos-fix-mbr-swap-removal
jprvita: It is used by `dd`.

eos-tech-support/eos-fix-mbr-swap-removal
jprvita: I agree with being less specific.

eos-tech-support/eos-fix-mbr-swap-removal
jprvita: I've added `echo` to get a list instead multiple lines from `lsblk`.

PR #545, opened 2018-12-13
eos-hack-reset-clubhouse: Reset cursor customizations

dnohales:
https://phabricator.endlessm.com/T24908

### DIFF
--- a/eos-tech-support/eos-hack-reset-clubhouse
+++ b/eos-tech-support/eos-hack-reset-clubhouse
@@ -18,4 +18,8 @@ rm -f ~/.var/app/com.endlessm.Fizzics/data/webkitgtk/localstorage/*
 echo "Resetting desktop icons"
 eos-reset-desktop
 
+echo "Resetting cursor"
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-hack-reset-clubhouse
```diff
@@ -18,4 +18,8 @@ rm -f ~/.var/app/com.endlessm.Fizzics/data/webkitgtk/localstorage/*
 echo "Resetting desktop icons"
 eos-reset-desktop
 
+echo "Resetting cursor"
+dconf reset /org/gnome/desktop/interface/cursor-size
+dconf reset /org/gnome/desktop/interface/cursor-theme
```
cosimoc: I think it's more customary to use `gsettings reset` here instead of `dconf` directly, since that works with any gsettings backend, not just the dconf one. Any reason not to use that here?

dnohales: No reason at all, it didn't come to my mind.

dnohales:
Updated.

PR #546, opened 2018-12-26
os-depends: ship xserver-xorg-input-wacom

cosimoc:
It's a small package and all that's needed to enable the GNOME Wacom
integration. Let's ship it in the core OS.

https://phabricator.endlessm.com/T14678

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -96,6 +96,7 @@ xauth
 xdg-user-dirs
 xdg-user-dirs-gtk
 xserver-xorg-input-libinput
+xserver-xorg-input-wacom
 # Some of these are really specific to x86, but just keep them off of arm for now
 xserver-xorg-video-amdgpu [!armhf]
 xserver-xorg-video-fbdev [!armhf]


wjt:
Quoth @jprvita and @ramcq on the associated ticket:

> IIRC we decided to leave xserver-xorg-input-wacom out of images on purpose when we switched to libinput, to support everything through libinput and on the assumption that it should be sufficient. While we (at the kernel team) never tested drawing tablets specifically, libinput does provide support for them, so my understanding is that they should be working without xserver-xorg-input-wacom: https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html

> Seconded, needs more investigation. GNOME must surely have a wacom story via libinput for Wayland, so why wouldn't it also work the same for xorg-input-libinput? libinput itself certainly has the functionality even if the config is different (and the xsetwacom tools/protocols no longer compatible). Could just as easily be a udev rules snafu causing libinput not to bind.

jprvita:
FWIW, wacoms were working when I last tested them on EOS 3.6.0 when I was in TW exactly 1 year ago. Only button remapping support was missing, which is tracked on T26581.

PR #549, opened 2019-01-22
tech-support/eos-convert-system*: fix journal dir perms

ramcq:
The eos-convert-system{,-qa} scripts have the same bug as described/seen in
https://phabricator.endlessm.com/T22687 which is that the journal dir
permissions are not correctly set after it is created, meaning that depending
on the execution environment of the script / setting of umask, some users (even
administrators) may be left unable to read the journal folder.


ramcq:
Noticed this whilst giving instructions to someone about how to enable persistent journal.

PR #550, opened 2019-01-29
eos-stage-flatpak: cope with remote names containing '.'

wjt:
`flatpak-builder --install`, among other tools, creates remotes of the
form "org.vim.Vim-origin". `ostree config get KEY` splits KEY on the
first '.' it contains, leading to:

    ++ ostree --repo=/var/lib/flatpak/repo config get 'remote "org.vim.Vim-origin".url'
    error: Key file does not have group “remote "org”

Since v2018.8, `ostree config get|set` accept a `--group` parameter,
precisely to deal with this case. Bump the version we depend on, and use
it.

https://phabricator.endlessm.com/T25253

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -199,7 +199,10 @@ Description: EndlessOS Kernel Team Development Tools
 
 Package: eos-tech-support
 Architecture: all
-Depends: ${misc:Depends}, ${eos:Depends}, usbutils
+Depends: ${eos:Depends},
+         ${misc:Depends},
+         ostree (>= 2018.8),
+         usbutils,
 Description: EndlessOS Technical Support Tools
  This is a collection of technical support tools that are installed by default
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -199,7 +199,10 @@ Description: EndlessOS Kernel Team Development Tools
 
 Package: eos-tech-support
 Architecture: all
-Depends: ${misc:Depends}, ${eos:Depends}, usbutils
+Depends: ${misc:Depends},
+         ${eos:Depends},
+         usbutils,
+         ostree (>= 2018.8),
```
pwithnall: Nitpick: One of my favourite smcvisms is for package lists in `debian/control` to be in alphabetical order.

wjt: I ran `:sort` on this and squashed the result in.

PR #554, opened 2019-03-15
core-depends: add podman and recommended packages

ramcq:
Add podman to the OS image, including slirp4netns and uidmap which are required
for using podman as an unprivileged user.

https://phabricator.endlessm.com/T25542

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -101,6 +101,8 @@ fonts-vlgothic
 fonts-wqy-zenhei
 foomatic-db-compressed-ppds
 fprintd
+# For rootless podman
+fuse-overlayfs
 gdb
 gedit
 # For showing the keyboard layout from the control center
@@ -172,6 +174,7 @@ openprinting-ppds
 openssh-server
 # For flatpak-builder
 patch
+podman
 # For modem support
 ppp
 printer-driver-all-enforce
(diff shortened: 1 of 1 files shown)


eos-core-depends
```diff
@@ -172,6 +172,7 @@ openprinting-ppds
 openssh-server
 # For flatpak-builder
 patch
+podman
```
wjt: Won't you also want to include `podman-docker` once that's in the distro?

wjt: Resolution: no, because without `docker-compose` support, having `/usr/bin/docker` provides false hope.

wjt:
How about fuse-overlayfs?

ramcq:
> How about fuse-overlayfs?

Voila. Ready to roll.

wjt:
It'd be nice if this package had enough of a build system that it could check if all the packages are actually present in the distro, eh?

eos-core-depends
```diff
@@ -101,6 +101,7 @@ fonts-vlgothic
 fonts-wqy-zenhei
 foomatic-db-compressed-ppds
 fprintd
+fuse-overlayfs
```
wjt: I think you want `# For rootless podman` here?

ramcq: Done!

PR #555, opened 2019-03-25
Check that deps exist at build time

wjt:
This would have prevented me from merging https://github.com/endlessm/eos-meta/pull/554 when fuse-overlayfs was not in the distro.

There are also some other fixes in here for things that lintian complained about. Many of them were harmless but one was real.

https://phabricator.endlessm.com/T26007

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -4,7 +4,9 @@ Priority: optional
 Maintainer: EndlessM Maintainers <maintainers@endlessm.com>
 Uploaders: Sjoerd Simons <sjoerd.simons@collabora.co.uk>, Hector Oron <hector.oron@collabora.co.uk>, Srdjan Grubor <sgnn7@sgnn7.org>
 Standards-Version: 3.9.3
-Build-Depends: debhelper (>= 9)
+Build-Depends: debhelper (>= 9),
+               libdpkg-perl,
+               meson,
 Homepage: http://www.endlessm.com
 
 Package: eos-core
--- a/debian/eos-tech-support.install
+++ b/debian/eos-tech-support.install
@@ -1,1 +1,1 @@
-eos-tech-support/* usr/bin
+eos-tech-support/eos-* usr/bin
--- a/eos-dev-recommends
+++ b/eos-dev-recommends
@@ -1,2 +1,1 @@
-eos-dev-config
 synergy
--- a/eos-metapackage
+++ b/eos-metapackage
@@ -10,10 +10,13 @@ eos-metapackage - create metapackages from depends files
 use strict;
 
 use Debian::Debhelper::Dh_Lib;
+use Dpkg::Deps qw( deps_parse deps_iterate );
 
 =head1 SYNOPSIS
 
 B<eos-metapackage> [S<B<debhelper options>>]
+
+B<eos-metapackage --check-params>
 
 =head1 DESCRIPTION
 
@@ -30,6 +33,10 @@ The format of the dependency files is one package per line. Lines
 beginning with "#" are skipped. Lines beginning with "include" will read
 in the referenced file.
 
+If run with C<--check-params>, rather than populating substvar files, the
+program will check whether all external packages for the current
+architecture listed in dependency files are installable.
+
 =head1 EXAMPLES
 
 eos-metapackage must be called before C<dh_installdeb>.
@@ -40,7 +47,10 @@ eos-metapackage must be called before C<dh_installdeb>.
 
 =cut
 
-init();
+my $CheckDeps = 0;
+init(options => {
+	"check-deps" => \$CheckDeps,
+});
 
 sub get_depends {
 	my $name=shift;
@@ -71,16 +81,52 @@ sub get_depends {
 	return @deps;
 }
 
+sub check_installable {
+	my ($depfile, $deps) = @_;
+	my @cmd;
+	my %options = (
+		# ignore packages for foreign architectures; they may not be
+		# installable on this architecture, and that's OK.
+		reduce_restrictions => 1,
+	);
+
+	verbose_print("checking all packages in $depfile are installable");
+	deps_iterate(deps_parse($deps, %options), sub {
+		my ($d) = @_;
+		my $package = $d->{package};
+
+		if (grep { $package eq $_ } @{$dh{DOPACKAGES}}) {
+			verbose_print("skipping '$package' which comes from this project");
+		} else {
+			push @cmd, $package;
+		}
+
+		return 1; # keep iteratin'
+	});
+
+	if (@cmd > 0) {
+		unshift(@cmd, ("apt-get", "install", "--download-only", "--dry-run", "--quiet"));
+		doit(@cmd);
+	}
+}
+
 sub subst_depends {
 	my $package=shift;
 	my $type=shift;
 	my $var=shift;
 	my $depfile="$package-$type";
 	my @deps;
+	my $deps;
 
 	# Allow the toplevel file to be missing (e.g., -recommends)
 	@deps = get_depends("$depfile") if -e "$depfile";
-	addsubstvar($package, $var, join(', ', @deps));
+	$deps = join(', ', @deps);
+
+	if ($CheckDeps) {
+		check_installable($depfile, $deps);
+	} else {
+		addsubstvar($package, $var, $deps);
+	}
 }
 
 foreach my $package (@{$dh{DOPACKAGES}}) {
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,14 @@
+project('eos-meta')
+
+if get_option('check-deps')
+  test('check-deps',
+    find_program('eos-metapackage'),
+    workdir : meson.current_source_dir(),
+    args : [
+      '--check-deps',
+    ],
+    env : [
+      'DH_VERBOSE=1',
+    ],
+  )
+endif
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ debian/eos-core
 debian/eos-dev
 debian/eos-dev-kernel
 debian/eos-dev-kernel-amd64
+obj-*
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 *.swp
 *~
 debian/*substvars
+debian/.debhelper
 debian/eos-apps
 debian/eos-core
 debian/eos-dev
+debian/eos-dev-kernel
+debian/eos-dev-kernel-amd64
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 *.log
-debian/*substvars
-debian/eos-core
-debian/eos-apps
-debian/eos-dev
 *.swp
 *~
+debian/*substvars
+debian/eos-apps
+debian/eos-core
+debian/eos-dev
--- a/debian/eos-tech-support.postinst
+++ b/debian/eos-tech-support.postinst
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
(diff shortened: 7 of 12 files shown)


wjt:
This is not done, but I'm requesting review as a pre-emptive “does this make sense?” check.

wjt:
OK, I think this actually _is_ ready. It fails for legit reasons at the moment, and also because `eos-dev-config` is not installable:

```
$ sudo apt install eos-dev-config 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 eos-dev-config : Depends: eos-node-modules-dev (>= 1.1.0) but it is not going to be installed
                  Depends: python3-difflint but it is not installable
E: Unable to correct problems, you have held broken packages.
```

Hooray, I suppose I get to go and fix that too.

wjt:
retest this please

wjt:
The `fuse-overlayfs` and `podman` issues are gone. I'd expect the remaining one to be fixed by merging https://github.com/endlessm/eos-dev-config/pull/68 .

dbnicholson:
> OK, I think this actually _is_ ready. It fails for legit reasons at the moment, and also because `eos-dev-config` is not installable:

I suggest you not actually install `eos-dev-config`. It does crazy stuff to your current user and running system.

https://github.com/endlessm/eos-dev-config/blob/debian-master/debian/postinst
https://github.com/endlessm/eos-dev-config/blob/debian-master/debian/gsettings-override
https://github.com/endlessm/eos-dev-config/blob/master/apt_conf/99allow-unauthenticated

Jasper used to refer to this package as `eos-bone-my-system`. If you're going to go changing it, I suggest tearing out about 90% of it. Basically anything that touches actual configuration should be at best an example shipped in the documentation.

wjt:
> I suggest you not actually install `eos-dev-config`. It does crazy stuff to your current user and running system.

I found that out the hard way!

> If you're going to go changing it, I suggest tearing out about 90% of it.

I don't really want to spend a lot of time on it. I'd be happy to remove it from `eos-dev-recommends`; or to go further and delete it from OBS and Jenkins and remove its packaging branch. (People can just run the scripts from from a Git checkout.) My only interest in it is that it breaks my lovely check here.

eos-metapackage
```diff
@@ -80,7 +80,37 @@ sub subst_depends {
...
 	# Allow the toplevel file to be missing (e.g., -recommends)
 	@deps = get_depends("$depfile") if -e "$depfile";
-	addsubstvar($package, $var, join(', ', @deps));
+
+	if ($ENV{EOS_META_CHECK}) {
+		verbose_print("checking all packages in $depfile are installable");
+		my @cmd;
+		my $arch = hostarch();
+
+		foreach my $dep (@deps) {
```
dbnicholson: I've never tried it, but you could try using `Dpkg::Deps::dep_parse($dep, reduce_restrictions => 1)`. You'd still have to filter out dependencies from this package, but it would allow you to cleanly manage the arch specific dependencies. See `dpkg-checkbuilddeps` and `dpkg-shlibdeps` for examples. You'd probably want to add a build dependency on `libdpkg-perl` even though it's implicitly pulled in by `debhelper`.

That said, if you've had enough of this I think it's probably fine to assume our simple usage matches the parsing below.

meson.build
```diff
@@ -0,0 +1,10 @@
+project('eos-meta')
```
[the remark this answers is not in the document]
wjt: I eyeballed it and decided it was fine, but you inspired me to check more closely and there are some slight differences between the `ninja dist` tarball and the one in OBS:

* Meson's `ninja dist` target generates a `.tar.xz`, whereas our fallback uses `.tar.gz`
* Files in Meson's `.tar.xz` have prefix `eos-meta/` whereas our fallback uses (eg) `eos-meta-1.5.0+dev1483.6470161/`
* I unpacked the two tarballs and ran `org.diffoscope.Diffoscope --exclude-directory-metadata --no-default-limits --max-page-size 40960000  --html eos-meta.html old/eos-meta-1.5.0+dev1483.6470161 new/eos-meta` on them. [Here's the output](https://willthompson.co.uk/misc/eos-meta.html). Interesting stuff:
    * `debian/source/options` is `--compression=gzip` in the tarball on OBS, and missing in Meson's; `debian/source/format` is `1.0` in the OBS tarball and `3.0 (native)` in the `meson dist` tarball.
    * All other changes are expected.

That said, I think all of the unexpected changes are actually things that `ci-integration/chroot-templates/obs-package` adjusts anyway, so this is all a red herring and practically speaking it does do the same thing.

debian/eos-tech-support.install
```diff
@@ -1 +1 @@
-eos-tech-support/* usr/bin
+eos-tech-support/eos-* usr/bin
```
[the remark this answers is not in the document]
wjt: Indeed!

eos-metapackage
wjt: Something was actually wrong in my hand-rolled parsing – I caught it excluding a `[!armhf]` package on my `amd64` system. It was pretty easy to swap in `Dpkg::Deps::deps_parse` and it made the surrounding code simpler. I also worked out how to add a command-line parameter rather than relying on a magic environment variable.

meson.build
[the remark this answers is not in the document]
wjt: I'm just running `jenkins-obs` locally. Once it's finished and OBS has build the results from my home project, I'll diff the `.deb`s.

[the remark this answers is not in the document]
dbnicholson: Yeah, I'm pretty sure it will be fine. The `obs-package` helper hacks the source package in the expected way regardless of what's in the original tarball.

eos-dev-recommends
```diff
@@ -1,2 +1 @@
-eos-dev-config
```
dbnicholson: I actually think we can make it useful by stripping out all the nasty parts, but I don't think it's worth the effort right now.

I do have a pipe dream to restore the `eosdev` ostree that uses the `eos-dev` metapackage and make `eosdev` master images for developers that have useful development utilities on them. At least for me a standard `eos` image is not capable enough for my work.

eos-metapackage
```diff
@@ -71,16 +81,52 @@ sub get_depends {
...
+	my ($depfile, $deps) = @_;
+	my @cmd;
+	my %options = (
+		# ignore packages for foreign architectures; they may not be
+		# installable on this architecture, and that's OK.
+		reduce_restrictions => 1,
+	);
+
+	verbose_print("checking all packages in $depfile are installable");
+	deps_iterate(deps_parse($deps, %options), sub {
```
dbnicholson: Fancy

eos-metapackage
```diff
@@ -71,16 +81,52 @@ sub get_depends {
...
+			verbose_print("skipping '$package' which comes from this project");
+		} else {
+			push @cmd, $package;
+		}
+
+		return 1; # keep iteratin'
+	});
+
+	if (@cmd > 0) {
+		unshift(@cmd, ("apt-get", "install", "--download-only", "--dry-run", "--quiet"));
```
wjt: This just doesn't work at all when building in my OBS branch. https://obs-master.endlessm-sf.com/build/home:wjt:branches:eos:master:endless/endless/x86_64/eos-meta/_log#

```
[  118s] DH_VERBOSE='1' /usr/src/packages/BUILD/eos-metapackage --check-params
[  118s] --- Listing only the last 100 lines from a long log. ---
[  118s] E: Unable to locate package gedit
[  118s] E: Unable to locate package gkbd-capplet
[  118s] E: Unable to locate package gnome-calculator
[  118s] E: Unable to locate package gnome-clocks
[  118s] E: Unable to locate package gnome-contacts
[  118s] E: Unable to locate package gnome-control-center
...
```

Presumably because there's no apt cache in the build chroot (and there's no network so I can't just chuck an `apt-get update` in there). So… I want the build to behave differently under `jenkins-obs` and `obs`, and I think the easiest way to achieve this is a build option that is enabled by default and disabled in `debian/rules`.

debian/control
```diff
@@ -4,7 +4,8 @@ Priority: optional
 Maintainer: EndlessM Maintainers <maintainers@endlessm.com>
 Uploaders: Sjoerd Simons <sjoerd.simons@collabora.co.uk>, Hector Oron <hector.oron@collabora.co.uk>, Srdjan Grubor <sgnn7@sgnn7.org>
 Standards-Version: 3.9.3
-Build-Depends: debhelper (>= 9)
+Build-Depends: debhelper (>= 9),
+               meson,
```
dbnicholson: I think you should include `libdpkg-perl` here even though it's an implicit dependency of `debhelper`.

eos-dev-recommends
wjt: Yep, I'm keen to not get rid of the `eos-dev` metapackage. It would also be a good base for a container image that could be run with `podman`.

eos-metapackage
dbnicholson: Yeah, it will be difficult to make this work in OBS. There's a repo at `/.build.binaries`, but it's constructed per build and only includes the packages OBS determines are needed for the build. It's not a full repo and certainly not a representation of our shared repo.

eos-metapackage
wjt: It's actually the easiest way to iterate this thing, AFAIK, because `deps_parse` always gives you back a collection of deps, even if it's a degenerate AND of one dep. It would also cope if we ever did introduce conditional deps `x | y`.

eos-metapackage
wjt: I “solved” this by adding an on-by-default option in `meson_options.txt`, and disabling it in `debian/rules`. This does of course mean that local builds with `dpkg-buildpackage -b` won't check this, but the goal is to check PRs, really.

wjt:
OK! This builds on both Jenkins and on OBS for both architectures. I ran Diffoscope to compare the `eos:master:endless` `.deb`s with the ones on my OBS branch. https://phabricator.endlessm.com/P625 – lots of timestamps are different, which is a bit annoying, but if one looks past that it seems that there are no unexpected changes.

eos-metapackage
dbnicholson: That works for now. I was thinking about this a little more, though. One issue with only running it in jenkins is that it will only check amd64 and not armhf. It would be ideal to have this run in OBS on all architectures.

How could you run this in OBS? Setup a temporary apt configuration like `chdist` does (or we've done 8 million times with `python-apt`). If you look at `dist_create` in `chdist`, you can see how it does this. Then you just run `apt` overriding the `APT_CONFIG` environment variable (or you could pass the `--config-file` option). One advantage of this approach is it points to an empty dpkg status file, so apt isn't influenced by the packages installed in the buildroot. You could do this yourself by making an empty temporary file and then setting the apt option `Dir::State::status` to it.

However, you'd still be a little stuck because you wouldn't know what dist to put in the sources file. Ideally you'd use something from `/etc/os-release` or `/etc/lsb-release` (assuming you have the proper build dep to pull them in). However, we don't actually fill those out with our codename correctly. We don't have `VERSION_CODENAME` in `/etc/os-release` and `DISTRIB_CODENAME` in `/etc/lsb-release` is just set to a generic `endless`. We'd want one of those to be `master` or `eosX.Y`. That would require a little rework of our branching process, though. Likewise, you don't know the components, but I think you could just hardcode `core endless` in there.

Anyways, food for thought but I think we can just go ahead with what you have as it's already a major improvement.

PR #557, opened 2019-04-01
eos-stage-flatpak: add a --remote-name option

cosimoc:
Optionally takes the name of a remote to stage, instead of all of
them at once.

https://phabricator.endlessm.com/T26115

### DIFF
--- a/eos-tech-support/eos-stage-flatpak
+++ b/eos-tech-support/eos-stage-flatpak
@@ -1,5 +1,10 @@
 #!/bin/bash -e
 
+ARGS=$(getopt -o r:h -l remote-name: -l help -n "$0" -- "$@")
+eval set -- "$ARGS"
+
+REMOTE_NAME=
+
 function usage {
     cat <<EOF
 Usage:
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-stage-flatpak
```diff
@@ -10,9 +15,31 @@ Arguments:
                 - demo: beta testing in advance of production releases
                 - prod: full production releases for general availability
     PASSWORD  Development password (for staging branch only)
+
+Options:
+    -r, --remote-name REMOTE_NAME  Name of the Flatpak remote to stage
```
pwithnall:
```suggestion
    -r, --remote-name REMOTE_NAME  Name of the flatpak remote to stage; if unspecified, all remotes configured in the system repository will be staged
```

eos-tech-support/eos-stage-flatpak
```diff
@@ -58,6 +85,10 @@ else
 fi
 
 for remote in `flatpak remote-list --system | awk {'print $1'}` ; do
+    if [ -n "$REMOTE_NAME" ] && [ "$remote" != "$REMOTE_NAME" ]; then
+        continue
+    fi
```
pwithnall: Actually, one other comment: some error handling should be added to warn the user if they specify a `--remote-name` which doesn’t actually exist (i.e. if they typo it), because otherwise this script will trivially succeed.

PR #558, opened 2019-04-02
T26120

cosimoc:
https://phabricator.endlessm.com/T26120

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -30,6 +30,7 @@ ARGS=$(getopt -o o:x:lp:r:nms:wL:fh \
     -l size: \
     -l fill \
     -l writable \
+    -l no-bios \
     -l label: \
     -l force \
     -l debug \
@@ -42,6 +43,7 @@ MBR=
 EXPAND=
 SIZE=
 WRITABLE=
+BIOS=true
 FORCE=
 PERSONALITY=base
 PRODUCT=eos
@@ -72,6 +74,7 @@ Options:
         --fill               Expand the image to fill the target disk
     -w, --writable           Allow image to be modified when run (which
                              prevents installing it later)
+        --no-bios            Do not create a BIOS boot partition
     -f, --force              don't ask to proceed before writing
     -h, --help               Show this message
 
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-write-live-image
```diff
@@ -389,6 +442,29 @@ fi
...
 
+if [ "$EXPAND" ]; then
+    if [ -n "$SIZE" ]; then
+        # Expand the image to the provided size
+        USED_SPACE_BYTES=$(df -P -B1 "$DIR_IMAGES_ENDLESS" | awk 'NR==2 {print $3}')
+        EXTRA_BYTES=$(($SIZE-$USED_SPACE_BYTES))
+
+        if [ $EXTRA_BYTES -lt 0 ]; then
+            echo "Cannot expand the image to $SIZE bytes, minimum size is $USED_SPACE_BYTES bytes"
+        fi
```
dbnicholson: This calculation seems a little funky to me. I think it's trying to avoid shrinking the image file, but it's looking at the currently used bytes in the enclosing filesystem. Those will roughly match, but I think comparing to the size of the image file itself would match better. Furthermore, this doesn't check if the requested size is bigger than the amount of free space. I think it could be a little better like this:
```
IMAGE_BYTES=$(du -b "${DIR_IMAGES_ENDLESS}/endless.img" | cut -f1)
FREE_SPACE_BYTES=$(df -P -B1 "$DIR_IMAGES_ENDLESS" | awk 'NR==2 {print $4}')
if [ -n "$SIZE" ]; then
  if [ $SIZE -lt $IMAGE_BYTES ]; then
    echo "Cannot expand the image to $SIZE bytes, minimum size is $IMAGE_BYTES bytes" >&2
    exit 1
  fi
  EXTRA_BYTES=$(($SIZE - $IMAGE_BYTES))
  if [ $EXTRA_BYTES -gt $FREE_SPACE_BYTES ]; then
    echo "Cannot expand the image to $SIZE bytes, only $FREE_SPACE_BYTES bytes available" >&2
    exit 1
  fi
else
  # Expand to fill all space
  SIZE=$(($IMAGE_BYTES + $FREE_SPACE_BYTES))
fi

if [ $SIZE -gt $IMAGE_BYTES ]; then
  echo "Expanding the image from $IMAGE_BYTES to $SIZE bytes"
  truncate --size "$SIZE" "${DIR_IMAGES_ENDLESS}/endless.img"
fi
```

cosimoc:
Thanks @dbnicholson! I updated the PR for your suggestions and it seems to work well.

PR #559, opened 2019-04-04
Drop packages dropped in the Great Buster Purge

wjt:
https://phabricator.endlessm.com/T26007

### DIFF
--- a/eos-dev-suggests
+++ b/eos-dev-suggests
@@ -21,7 +21,7 @@ gnome-tweak-tool
 gparted
 jq
 kbd
-linux-tools-3.10
+linux-tools-generic
 logrotate
 multistrap
 nmap
--- a/eos-dev-depends
+++ b/eos-dev-depends
@@ -27,7 +27,6 @@ emacs
 endless-ca-cert
 exuberant-ctags
 gdb
-gdebi
 git
 gitk
 gnome-common
--- a/eos-dev-recommends
+++ b/eos-dev-recommends
@@ -1,1 +1,0 @@
-synergy


wjt:
Yep, that's what I think too.

I'm a little confused because I thought that part of the point of `eos-dev-suggests` was to keep stuff in the distro that would otherwise be removed. Here's its opening comment:

```
# Packages listed here become Suggested by the eos-dev package.
#
# Such packages aren't installed by default when you install eos-dev.
# The main purpose of listing packages here is so that they aren't considered
# for deletion by scripts that look for unused packages in our distro. Such
# scripts have special provisions to consider anything listed here as a
# required part of our distro.
```

@dsd did your purge of pre-Buster packages (deliberately) ignore this?

eos-dev-suggests
```diff
@@ -21,7 +21,7 @@ gnome-tweak-tool
 gparted
 jq
 kbd
-linux-tools-3.10
+linux-tools-generic
```
dbnicholson: One other thought here - since `linux-tools-generic` comes with the `linux` package, this probably isn't needed since I highly doubt we'll accidentally remove `linux` :smile:

dsd:
> @dsd did your purge of pre-Buster packages (deliberately) ignore this?

I haven't actually done the proper purging yet. So far I've just removed a handful of packages such as those that fail to build, blocking soversion changes, would be painful to fix, no longer exist in Debian Buster, etc. In this case linux-tools-3.10 needed a rebuild as part of the upgrade process, and it was failing to build, not worth the effort keeping it around. In this case you can drop it from eos-dev-suggests.

The package purge script you're referring to has not been run recently; I will announce & run it when buster updates are closer to completion, it will delete around 300 packages.

dsd:
Pushed the discussed changes to fix the eos-meta build. Thanks

PR #560, opened 2019-04-05
eos-write-live-image: support windows tool for writable images

cosimoc:
Previously eos-write-live-image only wrote the windows tool to the
removable drive if a non-writable live was requested.

This commit changes the behavior so that if a windows tool was
explicitly provided to the script, it will be written regardless
of whether the image was writable or not, and preserves the existing
behavior otherwise.

https://phabricator.endlessm.com/T26120


wjt:
retest this please :robot:

PR #561, opened 2019-04-08
Add mousetweaks to eos-core-depends

GeorgesStavracas:
It is recommended by gnome-settings-daemon and, because
we do not install recommended packaged by default, it
needs to be explicitly included.

https://phabricator.endlessm.com/T26202

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -165,6 +165,8 @@ libpam-gnome-keyring
 # Assumed useful for apps that use libsasl2
 libsasl2-modules
 modemmanager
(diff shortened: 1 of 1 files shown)


eos-core-depends
```diff
@@ -165,6 +165,8 @@ libpam-gnome-keyring
 # Assumed useful for apps that use libsasl2
 libsasl2-modules
 modemmanager
+# For mouse accessibility options
```
wjt:
```suggestion
# For mouse accessibility options in gnome-settings-daemon
```

?

PR #562, opened 2019-04-09
eos-dev-unlock: pass args to ostree admin unlock

wjt:
It is useful in some situations (such as on a development system, or for
code that runs at boot) to install packages persistently on an
otherwise-unconverted system. `ostree admin unlock --hotfix` can be used
for this, but anecdotally can result in interesting behaviour if the
updaters are left running.

Support running `eos-dev-unlock --hotfix`, passing the argument through
to `ostree admin unlock`. The easiest way to achieve this is to just
pass all arguments (except `-h` or `--help`) through.


wjt:
@ManMower – it was your anecdote that I was thinking of.

PR #564, opened 2019-04-23
os-depends: add fscrypt and libpam-fscrypt to the ostree

ramcq:
https://phabricator.endlessm.com/T24742


ramcq:
test this please

PR #567, opened 2019-05-09
Add metapackage for eos-toolbox-dev

andrunko:
https://phabricator.endlessm.com/T26442

### DIFF
--- a/eos-toolbox-dev-depends
+++ b/eos-toolbox-dev-depends
@@ -1,0 +1,70 @@
+# The eos-toolbox-dev package is intended to be included in the eos-toolbox
+# container images as a way to provide the base development packages needed
+# for common development.
+#
+# Packages listed here should be commonly used, of high value for
+# development, and expected to be relevant to a whole group of Endless
+# developers.
+#
+# Some packages included in the eos-dev package (e.g. gnome dev packages)
+# are not included here on purpose given they increase the container image
+# size quite substantially. Developers that need the extra packages can
+# still install them manually using apt-get from an eos-toolbox container.
+
+acl
+apt-file
+apt-utils
+autoconf
+automake
+bash-completion
+buildah
+build-essential
+ca-certificates
+ccache
+curl
+dbus
+debootstrap
+devscripts
+diffutils
+endless-ca-cert
+findutils
+flatpak-xdg-utils
+gdb
+git
+gnupg
+hostname
+htop
+iftop
+iotop
+less
+libcap2-bin
+libnss-myhostname
+locales-all
+lsof
+make
+man-db
+manpages
+meson
+net-tools
+obs-build
+openssh-client
+osc
+passwd
+patchutils
+pbuilder
+pycodestyle
+podman
+procps
+python3
+quilt
+rsync
+strace
+sudo
+time
+traceroute
+unzip
+util-linux
+wget
+xzdec
+xz-utils
+zip
--- a/debian/control
+++ b/debian/control
@@ -210,3 +210,10 @@ Description: EndlessOS Technical Support Tools
  with the EndlessOS operating system.  They are not required as part of the
  normal user experience but are relatively small and helpful for advanced
  troubleshooting.
+
+Package: eos-toolbox-dev
+Architecture: all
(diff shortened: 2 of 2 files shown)


wjt:
retest this please, o great jenkins

andrunko:
> retest this please, o great jenkins

hehe, seems to be passing on jenkins now :).

andrunko:
Updated commit to sort the package list alphabetically :).

eos-toolbox-dev-depends
```diff
@@ -0,0 +1,74 @@
...
+# developers.
+#
+# Some packages included in the eos-dev package (e.g. gnome dev packages)
+# are not included here on purpose given they increase the container image
+# size quite substantially. Developers that need the extra packages can
+# still install them manually using apt-get from an eos-toolbox container.
+
+acl
+apt-file
+apt-transport-https
```
wjt:
```
wjt@tensionsheet-redux:~$ apt show apt-transport-https
...
Description: transitional package for https support
 This is a dummy transitional package - https support has been moved into
 the apt package in 1.5. It can be safely removed.
```

```suggestion
```

eos-toolbox-dev-depends
```diff
@@ -0,0 +1,74 @@
...
+apt-utils
+autoconf
+automake
+bash-completion
+buildah
+build-essential
+bzip2
```
wjt: I think this is a transitive dependency of build-essential – could it be removed?

eos-toolbox-dev-depends
```diff
@@ -0,0 +1,74 @@
...
+obs-build
+openssh-client
+osc
+passwd
+patchutils
+pbuilder
+pep8
+podman
+procps
+python
```
wjt: Must we explicitly pull in Python 2? It's a dependency of `osc` but at some point one would hope this would go away, and we would no longer need Python 2 for package builds. It's already been removed from the OS as distributed, though I know it's still a moderately widely-used build dep.

eos-toolbox-dev-depends
```diff
@@ -0,0 +1,74 @@
...
+manpages
+meson
+net-tools
+obs-build
+openssh-client
+osc
+passwd
+patchutils
+pbuilder
+pep8
```
wjt:
```
Description: Python PEP 8 code style checker - transitional package
 pep8 script was renamed to pycodestyle.
 .
 This is transitional package which provides symlink for pep8 script
```

Shall we ship `pycodestyle` directly instead?
```suggestion
pycodestyle
```

eos-toolbox-dev-depends
```diff
@@ -0,0 +1,74 @@
...
+python3
+quilt
+rsync
+strace
+sudo
+time
+traceroute
+unzip
+util-linux
+vim
```
wjt: Well, maybe. I actually do use Vim, but I have weaned myself onto the Flathub package of it. Is this needed in the countainer?

andrunko:
thanks @wjt - I will have a look at the comments after lunch.

eos-toolbox-dev-depends
andrunko: ack, I will remove it explicitly but I am afraid of relying on other pkg deps in general but this one likely wont go away ever :).

eos-toolbox-dev-depends
andrunko: ack, will remove it

eos-toolbox-dev-depends
andrunko: I thought we had some projects/components that would still use python2, but I am Ok with removing it explicitly, will do that.

eos-toolbox-dev-depends
andrunko: ack

eos-toolbox-dev-depends
andrunko: hehe, I may be biased here cause I only use vim as editor :). I would like to have a better editor than the default "vi" in the image but am Ok also to let the developer install whatever editor they want if you think that is an issue.

andrunko:
@wjt commit updated, please have a look again.

PR #569, opened 2019-05-22
eos-toolbox-dev: Remove buildah

andrunko:
buildah was getting installed as a dep for the eos-toolbox-dev
metapackage in order to allow running toolbox inside eos-toolbox
containers.
Now that toolbox (0.0.10) doesn't depend on buildah anymore, lets
remove the dep as not needed anymore.

https://phabricator.endlessm.com/T26430

Signed-off-by: Andre Moreira Magalhaes <andre@endlessm.com>

### DIFF
--- a/eos-toolbox-dev-depends
+++ b/eos-toolbox-dev-depends
@@ -17,7 +17,6 @@ apt-utils
 autoconf
 automake
 bash-completion
-buildah
 build-essential
 ca-certificates
 ccache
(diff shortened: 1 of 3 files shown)


eos-toolbox-dev-depends
```diff
@@ -17,7 +17,6 @@ apt-utils
 autoconf
 automake
 bash-completion
-buildah
```
wjt: Let's remove it from `eos-core-depends` as well since AIUI it's not needed any more? We can always add it back later.

andrunko: Done, updated - also removed from eos-dev-suggests.

PR #579, opened 2019-07-16
eos-enable-debian-apt: support -h and --help

wjt:
The easiest way to implement this is to treat an argument beginning with
'-' as being a request for help, since no Debian distros begin with 
hyphens.

This seems preferable to configuring a repo for the Debian distro
'--help', as I just did.

I also fixed a couple of warnings from shellcheck.

### DIFF
--- a/eos-tech-support/eos-enable-debian-apt
+++ b/eos-tech-support/eos-enable-debian-apt
@@ -16,7 +16,7 @@ echo "Additionally, apt will only work after you have taken the usual steps of"
 echo "making it available (e.g. eos-convert-system, eos-dev-unlock, etc)"
 echo
 
-if [[ $# -lt 1 ]] || [[ $1 == -* ]]; then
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-enable-debian-apt
```diff
@@ -16,7 +16,7 @@ echo "Additionally, apt will only work after you have taken the usual steps of"
 echo "making it available (e.g. eos-convert-system, eos-dev-unlock, etc)"
 echo
 
-if [[ $# < 1 ]]; then
+if [[ $# -lt 1 ]] || [[ $1 == -* ]]; then
```
andrunko: Changes lgtm but I would probably change this conditional to `if [[ $# -ne 1 ]] || [[ $1 == -* ]];` to make sure only a single argument is allowed - although we only use the first argument so I will leave it at your discretion...

wjt: Pushed this and one other cosmetic change; how does it look?

PR #581, opened 2019-07-19
os-depends: install spice-webdavd

wjt:
This (partially) enables shared folder support when running Endless OS
as a guest in Boxes, the only VM tool available on Endless OS. It adds
just 48KB (since we already ship its dependencies) and is conditionally
enabled by a udev rule which detects support for SPICE's WebDAV channel
at boot time.

https://phabricator.endlessm.com/T27317


wjt:
@treitter I think this would unblock your development workflow.

PR #585, opened 2019-07-31
eos-initramfs-efi: Add eos-plymouth-theme

ManMower:
I broke the kernel build because plymouth isn't present when dracut
is invoked to make the efi initramfs+kernel blobs.

Adding eos-plymouth-themes here brings in several missing dependencies:
plymouth, plymouth-themes, eos-plymouth-theme-logos

https://phabricator.endlessm.com/T27042

### DIFF
--- a/eos-initramfs-efi-depends
+++ b/eos-initramfs-efi-depends
@@ -1,3 +1,6 @@
+eos-plymouth-theme
+plymouth
+e2fsprogs
 eos-boot-helper
 dracut
 ostree


eos-initramfs-efi-depends
```diff
@@ -1,3 +1,4 @@
+eos-plymouth-theme
```
dbnicholson: Please also explicitly add `plymouth` as that's what the dracut module actually uses. In fact, looking at `/etc/dracut.conf.d/endless.conf`, I see other things in `dracutmodules` that are not explicitly called out here. E.g., `systemd` and `e2fsprogs`.

ManMower: systemd is already in there, but I was counting on dracut's existing dependency on e2fsprogs to fill that need, much in the same way I was counting on eos-plymouth-theme to bring in plymouth.

I've force pushed changes adding both plymouth and e2fsprogs to be more direct about dependencies.  I hope I've got them all now. :/

PR #596, opened 2019-09-10
Add git-buildpackage to eos-dev-depends

GeorgesStavracas:
It's the tool that the Debian GNOME team uses, but it's
not a dependency of any other Debian tool.

https://phabricator.endlessm.com/T27741


andrunko:
bot, retest this please.

PR #597, opened 2019-09-16
eos-core-depends: Add efitools for PAYG secure boot key management

ManMower:
We need efitools to make authenticated efi variable updates for
managing secure boot keys (PK, KEK, db) on PAYG systems.

https://phabricator.endlessm.com/T27106

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -38,6 +38,8 @@ dnsmasq-base
 # For brasero video DVD functionality
 dvdauthor
 e2fsprogs-l10n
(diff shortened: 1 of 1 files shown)


ManMower:
Was unsure if the new dep should go in eos-core-depends or os-depends or somewhere else entirely...

jprvita:
os-depends gets included by the installer as well. I believe eos-core is indeed the right place (there is a new eos-core-amd64 that could work iiuc, but it is not really being used yet).

PR #607, opened 2019-12-18
eos-diagnostics: include 'zramctl' output

wjt:
On my system, this looks like:

```
NAME       DISKSIZE  DATA COMPR ALGORITHM STREAMS ZERO-PAGES TOTAL MEM-LIMIT MEM-USED MIGRATED MOUNTPOINT
/dev/zram0      23G 51.8M  4.1M lzo-rle         8       7765  7.6M        0B   192.5M      30B [SWAP]
```

I think it may be helpful for interpreting the RAM and "swap" usage
figures from 'free -mh'.

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -562,7 +562,8 @@ let diagnostics = [
         hardwareInfo: true,
         content: function() {
             return trySpawn('free -mh') + '\n' +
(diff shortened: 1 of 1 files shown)


jprvita:
It would be nice to push this to eos3.7 as well, assuming the command also exists there.

wjt:
It does. Pushed as 49553ae39323b59963b8bda9824ea9c83ce6d3f6

PR #612, opened 2020-03-05
Include the eos-watermark-extension package by default

GeorgesStavracas:
This feature is being moved from GNOME Shell to an extension
shipped by default.

Notice that this depends on https://github.com/endlessm/eos-build/pull/1090 -- the package doesn't exist yet.

https://phabricator.endlessm.com/T28778


andrunko:
retest this please

PR #613, opened 2020-03-06
Include the eos-hack-extension package by default

danigm:
This feature is being moved from GNOME Shell to an extension
shipped by default.

https://phabricator.endlessm.com/T28568


GeorgesStavracas:
retest this please

PR #616, opened 2020-03-16
Remove eos-launch-all script

wjt:
This has not worked since 3.4.0: gtk-launch is no longer installed. It
comes from the libgtk-3-bin package, which we removed as part of
https://phabricator.endlessm.com/T5124. Evidently nobody wants to launch
all apps and see if they work any more.

(The difference between gtk-launch and just running the app directly is
that the former tells the shell about it, which means the speedwagon can
show up.)


andrunko:
LGTM;

PR #617, opened 2020-03-17
eos-write-live-image: Allow provisioning of persistent storage space

ManMower:
We now have a command line option to fill the eoslive partition with
a file for persistent storage space.

Currently this is not a fast operation - truncating the file to size
appears to be slow, yet not quite as slow as zero filling the space
would be.  It is hoped that this will be faster in the future with
a non-FUSE exfat driver.  In that case we can reconsider whether
this mode of operation should be default.

Untested on NTFS.

https://phabricator.endlessm.com/T14891

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -19,7 +19,7 @@ if [ ! -f "$EOS_DOWNLOAD_IMAGE" ]; then
     EOS_DOWNLOAD_IMAGE='eos-download-image'
 fi
 
-ARGS=$(getopt -o o:x:lp:r:nms:wL:fh \
+ARGS=$(getopt -o o:x:lp:r:nms:wL:fPh \
     -l os-image: \
     -l windows-tool: \
     -l latest \
@@ -34,6 +34,7 @@ ARGS=$(getopt -o o:x:lp:r:nms:wL:fh \
     -l label: \
     -l force \
     -l debug \
+    -l persistent \
     -l help \
     -n "$0" -- "$@")
 eval set -- "$ARGS"
@@ -50,6 +51,7 @@ PERSONALITY=base
 PRODUCT=eos
 LABEL=
 DEBUG=
+PERSISTENT=
 
 usage() {
     local SELF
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-write-live-image
```diff
@@ -484,6 +491,15 @@ if [ "$EXPAND" ]; then
     fi
 fi
 
+if [ "$PERSISTENT" ]; then
+    FREE_SPACE_BYTES=$(df -P -B1 "$DIR_IMAGES_ENDLESS" | awk 'NR==2 {print $4}')
```
wjt: What happens if this is not a multiple of 512? (I think that's the sector size device-mapper requires). Is it guaranteed to be a multiple of 512 thanks to some property of NTFS or exFAT?

ManMower: It appears that on exFAT the free space is always* a multiple of some larger power of 2 than 512.

* I did a couple of random tests.

Being a FAT based filesystem this is probably always going to be true, but I'm just waving my hands.  Would you prefer explicitly enforcing a 512 byte alignment just in case?

wjt: A comment to this effect is probably OK! Or you could divide by 1024 and specify the unit as K when calling `truncate`. Probably overengineering though.

ManMower: I've pushed changes to do the division thing... (I originally wanted to use df -B1024, but it rounds up)

eos-tech-support/eos-write-live-image
```diff
@@ -484,6 +491,17 @@ if [ "$EXPAND" ]; then
     fi
 fi
 
+if [ "$PERSISTENT" ]; then
+    # we need 512 byte alignment for device mapper, so let's do this in KBytes.
+    FREE_SPACE_KBYTES=$(df -P -B1 "$DIR_IMAGES_ENDLESS" | awk 'NR==2 {print $4}')
+    let FREE_SPACE_KBYTES=FREE_SPACE_KBYTES/1024
```
wjt: Wow, I have *never* seen `let` before. I have always written this kind of thing as:
```bash
FREE_SPACE_KBYTES=$(( FREE_SPACE_KBYTES / 1024 ))
```

PR #623, opened 2020-06-18
eos-enable-coredumps: Update comment at the top

mwleeds:


eos-tech-support/eos-enable-coredumps
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -e
 #
-# Enables coredumps and adjusts storage path to work with coredumpcl
+# Enables coredumps and adjusts storage path to work with coredumpctl.
+# This takes effect immediately without requiring a reboot.
```
jprvita, line 4: Good suggestion! Perhaps we should also note that this persists over reboot, and no crash metrics are collected once this is enabled?

mwleeds: Also added the path where the coredumps are found.

PR #627, opened 2020-07-30
eos-write-live-image: add --copy-path option

dnohales:
This option allows to copy extra files to the USB exFAT partition.

https://phabricator.endlessm.com/T30586


jprvita:
Ah, and please add a link to the phabricator task on the commit message as well, it helps a lot when trying to track where changes come from.

dnohales:
Thanks @jprvita, updated.

PR #629, opened 2020-08-04
core: Add eos-panel-extension

andrunko:
It will replace the panel modifications we used to carry on GNOME Shell itself.

https://phabricator.endlessm.com/T30445


andrunko:
retest this please

PR #632, opened 2020-08-07
Add eos-usb-internet-writer command

dnohales:
Leaving this on WIP until we fix all the TODOs, they are in the script itself, but repeating those here:

- [ ] Use real URLs for Kolibri portable apps when LE have the official builds
- [ ] Add support for auto loading ZIM files from USB key in Kiwix for macOS
- [ ] Install the final Kolibri content when ready

https://phabricator.endlessm.com/T30586

### DIFF
--- a/eos-tech-support/eos-write-usb-internet
+++ b/eos-tech-support/eos-write-usb-internet
@@ -1,0 +1,260 @@
+#!/bin/bash -e
+set -o pipefail
+
+eos_write_live_image=$(dirname "$0")/eos-write-live-image
+if [ ! -f "$eos_write_live_image" ]; then
+    eos_write_live_image='eos-write-live-image'
+fi
+
+kiwix_windows_url=http://download.kiwix.org/release/kiwix-desktop/kiwix-desktop_windows_x64.zip
+kiwix_macos_url=https://download.kiwix.org/release/kiwix-desktop-macos/kiwix-desktop-macos.dmg
+
+# TODO: use real URLs when LE have the official builds
+kolibri_windows_url=http://eor-desktop:8000/Kolibri-0.14.0-win-signed.zip
+kolibri_macos_url=http://eor-desktop:8000/kolibri-macos.zip
+
+kiwix_windows_lnk_base64_data=$(cat <<BASE64
+TAAAAAEUAgAAAAAAwAAAAAAAAEbjAggAIAAAAKzpU1dkYNYBFChhV2Rg1gEUKGFXZGDWASiTRgAAAAAA
+AQAAAAAAAAAAAAAAAAAAAOsAFAAfUOBP0CDqOmkQotgIACswMJ0ZAC9DOlwAAAAAAAAAAAAAAAAAAAAA
+AAAAVgAxAAAAAAD7UCmkEABXaW5kb3dzAEAACQAEAO++c06sJPtQKaQuAAAACiMKAAAABQAAAAAAAAAA
+AAAAAAAAANWztQBXAGkAbgBkAG8AdwBzAAAAFgBmADIAKJNGAPZQe6EgAGV4cGxvcmVyLmV4ZQAASgAJ
+AAQA7772UHuh9lB7oS4AAADmvAsAAAAPAAAAAAAIAQAAAAAAAAAA7PhqAGUAeABwAGwAbwByAGUAcgAu
+AGUAeABlAAAAHAAAAEYAAAAcAAAAAQAAABwAAAAtAAAAAAAAAEUAAAARAAAAAwAAAJhEo1YQAAAAAEM6
+XFdpbmRvd3NcZXhwbG9yZXIuZXhlAAAiACIALgBrAGkAdwBpAHgALQB3AGkAbgBkAG8AdwBzAFwAawBp
+AHcAaQB4AC0AZABlAHMAawB0AG8AcAAuAGUAeABlACIAKwBIADoAXABjAG8AbgB0AGUAbgB0AFwALgBr
+AGkAdwBpAHgALQB3AGkAbgBkAG8AdwBzAFwAawBpAHcAaQB4AC0AZABlAHMAawB0AG8AcAAuAGUAeABl
+ABQDAAABAACgJXdpbmRpciVcZXhwbG9yZXIuZXhlAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAlAHcAaQBuAGQAaQByACUAXABlAHgAcABsAG8AcgBl
+AHIALgBlAHgAZQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-usb-internet-writer
```diff
@@ -0,0 +1,229 @@
...
+
+Arguments:
+    IMAGE_OPTIONS       Everything after -- is passed to the $eos_write_live_image
+                        command.
+
+Options:
+    -t, --tmp-dir PATH  Temporary directory to use for downloading files and
+                        save the USB key files to copy, user is in charge of
+                        cleaning this directory if required, the command will
+                        try to all the content from this directory.
```
jprvita, line 34: Looks like part of the phrase  went missing here.

eos-tech-support/eos-usb-internet-writer
```diff
@@ -0,0 +1,229 @@
...
+        --command=/app/libexec/kolibri \
+        org.learningequality.Kolibri "$@"
+}
+
+# Install required Flatpak apps
+apps_to_install=(
+    "com.endlessm.encyclopedia.en"
+    "org.learningequality.Kolibri"
+)
+echo "Installing apps ${apps_to_install[*]}. Make sure you have the Flatpak repo installed for those"
```
jprvita, line 166: I don't know if we should be installing apps to the host running this script. Maybe just check if they are installed and if not, print an error telling the user they should install / update these flatpaks before running the script.

eos-tech-support/eos-usb-internet-writer
```diff
@@ -0,0 +1,229 @@
...
+for app in "${apps_to_install[@]}"; do
+    flatpak install -y --or-update "$app"
+done
+
+# Prepare tmp directory
+mkdir -p "$tmp_dir/workspace"
+mkdir -p "$tmp_dir/downloads"
```
jprvita, line 173: Using macros / constants here would make the code a bit easier to read.

eos-tech-support/eos-usb-internet-writer
```diff
@@ -0,0 +1,229 @@
...
+
+# Download Software packages
+wget -c -O "$tmp_dir/downloads/kiwix-win.zip" http://download.kiwix.org/release/kiwix-desktop/kiwix-desktop_windows_x64.zip
+wget -c -O "$tmp_dir/downloads/kiwix-macos.dmg" https://download.kiwix.org/release/kiwix-desktop-macos/kiwix-desktop-macos.dmg
+
+# TODO: use real URLs when LE have the official builds
+wget -c -O "$tmp_dir/downloads/kolibri-win.zip" http://eor-desktop:8000/kolibri-win.zip
+wget -c -O "$tmp_dir/downloads/kolibri-macos.zip" http://eor-desktop:8000/kolibri-macos.zip
```
jprvita, line 181: Can we use macros / constants for these URLs as well? This also helps if we need to update them in the future, especially if we move all constants to the beginning of the file.

eos-tech-support/eos-usb-internet-writer
```diff
@@ -0,0 +1,229 @@
...
+run_kolibri manage importcontent network 9a46768f1ea24daa86ae693e4a04d04b
+
+# Write USB key
+extra_data=()
+while IFS= read -r -d '' path
+do
+    extra_data+=("--extra-data" "$path")
+done < <(find "$tmp_dir/workspace" -mindepth 1 -maxdepth 1 -print0)
+
+sudo $eos_write_live_image $debug "${extra_data[@]}" "$@"
```
jprvita, line 229: This script should probably be called with as root (or with `sudo`) instead of escalating privileges here, and we should check that similarly to how `eos-write-live-image` does it. Otherwise, if the calling user does not have admin access, we just wasted all the work up to this point.

Have you designed this so the `eos-write-live-image` arguments (like `--os-image`, `--persistent`, `--persistent-size` etc) are all passed after ` -- `? But it seems (unless I'm reading this wrong, which may be possible) you are also expecting a list of paths, which you then prepend each with `--extra-data`, is that it? If so, this seems a bit inconsistent and unexpected from the user's perspective. Also, whatever the approach is decided here, let's make it clear in the usage message.

(As a side note, I'm not a big fan of blindly forwarding command-line options like that in general, but I see the benefit of not having to parse and validate all of `eos-write-live-image` arguments, and keep them in sync between both scripts moving forward, so I think for the scope and intended usage of this tool this is the right approach.)

eos-tech-support/eos-usb-internet-writer
```diff
@@ -0,0 +1,229 @@
...
+
+Options:
+    -t, --tmp-dir PATH  Temporary directory to use for downloading files and
+                        save the USB key files to copy, user is in charge of
+                        cleaning this directory if required, the command will
+                        try to all the content from this directory.
+                        (default: $tmp_dir)
+        --debug         Enable debugging output
+    -h, --help          Show this message
+EOF
```
dylanmccall, line 38: I realize it's possible to pass it after --, but do you think it would be okay to add an eos_windows_launcher option here? Then this script would deal with the --windows-tool and --label options for eos-write-live-image on its own, like we do in the write-usb command for Hack. That way it could be a required argument. I think that makes sense since it's a key part of the eos-usb-internet thing, and then the options we add after -- would be strictly optional things related to the Endless OS image.

eos-tech-support/eos-usb-internet-writer
dnohales: I didn't want to run Kolibri and do the other processing stuff as root, I wasn't sure of the unseen implications of that so I played safe, maybe it's not that bad since I'm only sharing the Kolibri home folder to it.

That's exactly what it does, it prepends the `--extra-data` flags to copy the workspace filder and then it goes with everything after `--`, the flags after `--` could also include `--extra-data` if user wants, I could make it more clear in the usage message what's going on. Initially, my approach was to hardcode most of the flags passed to `eos-write-live-image` according to our USB specs and have a higher level flags for `eos-usb-internet-writer`. I decided later that this was maybe this was easier and more flexible

eos-tech-support/eos-usb-internet-writer
dnohales: You are right,

eos-tech-support/eos-usb-internet-writer
dnohales: All right, maybe as suggested here https://github.com/endlessm/eos-meta/pull/632#discussion_r468662709 it's better to create a higher level set of flags for this command and let this command figure out most of the flags for `eos-write-live-image`.

eos-tech-support/eos-usb-internet-writer
jprvita: Ok, makes sense not running the whole thing as root then. I wonder if there is a way we could check if the user has sudo access (I don't know if that is possible) in the beginning of the script. But I'm probably overthinking this, as the intended audience is Endless' devs. So yeah, let's keep this bit as is.

For the command line arguments, the direction you and @dylanmccall are discussing seems sensible to me.

dnohales:
Updated: Moved some variables to the top of the file for better organization, I'm also copying the full `com.endlessm.encyclopedia.en.Content` runtime directory to `eoslive/.kiwix` to facilitate the symlinking for Endless OS encyclopedia.

I didn't find a proper way to make flags part of this command since there's stuff we still don't know (where's the launcher to use, should we calculate the persistent image dynamically or should we hardcode it later, product and personality to use). I opted to leave everything like before and being a bit more clear in the help message, just for simplicity, flexibility and lack of knowledge on how to proceed for now.

eos-tech-support/eos-write-usb-internet
```diff
@@ -0,0 +1,243 @@
...
+ROAY70WqlJxDyEbD+ctTr0C20uoRnalc83CNzeJkKqtE4BjvRaqUnEPIRsP5y1OvQLbS6hGdqVzzcI3N
+4gAAAAA=
+BASE64
+)
+
+kiwix_library_xml=$(cat <<XML
+<library version="20110515">
+    <book
+        id="97e3f348-38f4-6787-b95d-1e4886bd803c"
+        path="../.kiwix/files/com.endlessm.encyclopedia.en-openzim-subscription/1.zim"
```
wjt, line 86: This will need to be in synch with what's expected in Endless OS / Flatpak land, c.f. https://github.com/endlessm/eos-boot-helper/pull/305

eos-tech-support/eos-write-usb-internet
```diff
@@ -0,0 +1,243 @@
...
+)
+
+args=$(getopt -o t:h \
+    -l tmp-dir: \
+    -l debug \
+    -l help \
+    -n "$0" -- "$@")
+eval set -- "$args"
+
+tmp_dir=/tmp/eos-usb-internet-writer-tmp
```
wjt, line 107: mktemp -d

eos-tech-support/eos-write-usb-internet
```diff
@@ -0,0 +1,243 @@
...
+            shift
+            break
+            ;;
+    esac
+done
+
+workspace_dir="$tmp_dir/workspace"
+downloads_dir="$tmp_dir/downloads"
+
+if ! hash apfs-fuse 2>/dev/null; then
```
wjt, line 160: Can we ship this in Endless OS?

eos-tech-support/eos-write-usb-internet
wjt: Is `/tmp` going to be big enough? On Endless OS it is a tmpfs, limited in size to the system's physical RAM.

wjt: (If not, `/var/tmp` is disk-backed on Endless OS)

dnohales: Take into account this is the default and it's configurable, but I can change the default, sure.

eos-tech-support/eos-write-usb-internet
jprvita: It looks like this is not packaged for Debian or Ubuntu already. Debian ships **libfsapfs-utils**, which includes a different fuse apfs implementation, `fpafsmount`. Would that work for what we are trying to achieve here? If so, that would add ~2MB to the ostree, so it should be ok to ship it, in which case we should have a separate commit in this same PR including it.

dnohales:
Updated:

* Added support for `libfsapfs-utils` (while leaving support for `apfs-fuse`), now `p7zip-full` is required, which is already available on Endless.
* Fix Kolibri home creation (failed to run Flatpak with `--filesystem` on a non existing directory, had to create it first).
* Added `provisiondevice.json` to Kolibri home.

Now I'm closing this and moving it to https://github.com/endlessm/eos-release-engineering/pull/35

PR #633, opened 2020-08-14
core: Add eos-tour-extension

danigm:
https://phabricator.endlessm.com/T30237


andrunko:
retest this please

danigm:
> The change itself is Ok but I think we should rename the package to eos-onboarding-extension (or rename the repo) as this got me confused for a minute until I checked the packaging.

I've renamed the package, so now it's `eos-onboarding-extension` and it has a coherent name with the repository.

PR #637, opened 2020-08-28
eos-write-live-image: Never create BIOS Boot partition

wjt:
Because Windows offers to format the BIOS Boot partition, we don't want to create it, hence `--no-bios`.  However, previously we didn't write a BIOS bootloader when `--no-bios` was passed, which means the resulting USB stick can only be booted on UEFI systems.

Even if we don't have an entry in the partition table, we can leave a gap where the partition ought to be, and write the bootloader there anyway.  This allows the resulting USB stick to be booted on a legacy BIOS system too.

Continue to accept `--no-bios`, because we have documentation on creating Hack Keys that mentions this parameter, but ignore it.

This **DOES NOT WORK**. 1 time out of 4 it booted in GNOME Boxes but Initial Setup didn't come up. All other times it has hung at `Booting from Hard Disk...`.

https://phabricator.endlessm.com/T30692

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -48,7 +48,6 @@ MBR=
 EXPAND=
 SIZE=
 WRITABLE=
-BIOS=true
 WINDOWS_TOOL_PROVIDED=
 EXTRA_DATA_PATHS=()
 FORCE=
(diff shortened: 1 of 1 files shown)


wjt:
This does not work for reasons outlined in the

PR #644, opened 2020-09-10
eos-write-live-image fixes for use on build server

wjt:
I ran `eos-write-live-image` on a huge VM the other day, and had to bash it a bit to make it work. All the changes seem OK so here they are.

https://phabricator.endlessm.com/T30765


jprvita (approved):
Looks correct and neat!

PR #645, opened 2020-09-11
eos-write-live-image: Never create BIOS Boot partition

wjt:
This is another attempt at https://github.com/endlessm/eos-meta/pull/637, with an additional change to reverse the partition order to avoid triggering the undesired behaviour in eos-boot-helper.

It is on top of #644 to avoid conflicts.

Once again it doesn't work, for unknown reasons.

https://phabricator.endlessm.com/T30692


dnohales:
I think this needs a rebase?

wjt:
It also doesn't work and I have abandoned this for now, see linked ticket

PR #646, opened 2020-09-15
eos-write-metrics-collector-usb: script to create collector USB

mschiu77:
Erase and reformat the plugged in USB storage device for offline
metrics data collecting.

### DIFF
--- a/eos-tech-support/eos-write-metrics-collector-usb
+++ b/eos-tech-support/eos-write-metrics-collector-usb
@@ -9,7 +9,7 @@
 # users to execute to collect/upload offline metrics
 # data.
 
-EOSMETRICS_LIBEXEC="/usr/share/libexec/eos-metrics"
+EOSMETRICS_LIBEXEC="/usr/lib/eos-event-recorder-daemon"
 
 USERID=$(id -u)
 if [ "$USERID" != "0" ]; then
@@ -85,9 +85,7 @@ fi
 
 echo "== Populating $PART =="
 TARGET_MOUNTPOINT="$(findmnt --noheadings --output TARGET --source $PART)"
-cp -v $EOSMETRICS_LIBEXEC/metrics.py "$TARGET_MOUNTPOINT"
 cp -v $EOSMETRICS_LIBEXEC/eos-metrics-collector.exe "$TARGET_MOUNTPOINT"
-cp -v $EOSMETRICS_LIBEXEC/eos-metrics-uploader.exe "$TARGET_MOUNTPOINT"
 
 echo "== Unmounting $PART =="
 udisksctl unmount -b "$PART" --no-user-interaction
--- a/eos-tech-support/eos-write-metrics-collector-usb
+++ b/eos-tech-support/eos-write-metrics-collector-usb
@@ -1,0 +1,95 @@
+#!/bin/bash -e
+
+# Detect the plugged in USB storage and prompt users
+# that the contents of the storage will be erased
+# and the whole disk will be reformatted to a single
+# FAT32 partition (dos).
+# Then copy both eos-metrics-collector and the
+# eos-metrics-uploader with .exe extension for the
+# users to execute to collect/upload offline metrics
+# data.
+
+EOSMETRICS_LIBEXEC="/usr/share/libexec/eos-metrics"
+
+USERID=$(id -u)
+if [ "$USERID" != "0" ]; then
+    echo "Program requires superuser privileges" >&2
+    exit 1
+fi
+
+USB_DISK_NUM=0
+for device in /sys/block/* ; do
+    if udevadm info --query=property --path=$device | grep -q ^ID_BUS=usb ; then
+        (( USB_DISK_NUM+=1 ))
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-write-metrics-collector-usb
```diff
@@ -0,0 +1,92 @@
...
+
+# Add the basic data partition
+PARTS="start=2048, name=EOSMETRICS, type=EBD0A0A2-B9E5-4433-87C0-68B6B72699C7
+$PARTS"
+
+# Reconstruct the table
+TABLE="$HEADER
+$PARTS"
+
+echo "$TABLE" | sfdisk --force --no-reread "$TARGET_DISK"
```
jprvita, line 72: Why are you using `--no-reread` here?

eos-tech-support/eos-write-metrics-collector-usb
```diff
@@ -0,0 +1,92 @@
...
+PART=$(sfdisk -d "$TARGET_DISK" | grep 'EOSMETRICS' | cut -f1 -d' ')
+
+mkfs.vfat -n EOSMETRICS "$PART"
+
+# Give udisks a chance to notice the new partition
+partprobe "$TARGET_DISK"
+udevadm settle
```
jprvita, line 79: We want to make sure the partition table is up-to-date before trying to write a filesystem to it, so we should move the `mkfs.vfat` after these partprobe / udevadm settle calls.

eos-tech-support/eos-write-metrics-collector-usb
```diff
@@ -0,0 +1,92 @@
...
+
+# Try to mount the fat32 partition
+udisksctl mount -b "$PART" --no-user-interaction || exit 1
```
jprvita, line 82: Let's report an error if mounting fails here.

eos-tech-support/eos-write-metrics-collector-usb
```diff
@@ -0,0 +1,92 @@
...
+# Grab the mount point
+TARGET_MOUNTPOINT=$(grep "$PART" /proc/mounts | cut -f2 -d' ')
+if test -n "$TARGET_MOUNTPOINT" -a -d "$TARGET_MOUNTPOINT"
+then
+    # Copy the script files here.
+    cp /lib/eos-metrics-collector.exe "$TARGET_MOUNTPOINT"/
+    cp /lib/bin/eos-metrics-uploader.exe "$TARGET_MOUNTPOINT"/
```
jprvita, line 89: We don't have a final location where these files are going to be shipped yet, since that code hasn't been merged, but I think it should be `/usr/libexec/eos-metrics/`, or something similar. I suggest using a constant for that directory name.

Also, if the architecture remains the same, you also need to ship the `metrics.py` file, otherwise these scripts will fail to import classes from there.

jprvita (changes requested):
Left a few comments that need addressing.

eos-tech-support/eos-write-metrics-collector-usb
mschiu77: I don't want check whether the disk is in use. Or maybe I can prompt a warning message before sfsidk?

jprvita:
bot, test this please

eos-tech-support/eos-write-metrics-collector-usb
jprvita: This `--no-reread` tell `sfdisk` to go ahead with the re-partitioning even if the disk is being used. You are already unmounting the disk before calling `sfdisk`, so it should not be in use -- if it is, something is wrong and we don't want `sfdisk` to proceed blindly.

eos-tech-support/eos-write-metrics-collector-usb
```diff
@@ -0,0 +1,107 @@
...
+    echo "Please plug in the USB storage device."
+    exit 1
+fi
+
+prompt_input
+
+TARGET_DISK=${DISK[$((USER_CHOICE-1))]}
+
+echo "disk name $TARGET_DISK"
+MOUNTPOINT=$(grep "$TARGET_DISK" /proc/mounts | cut -f2 -d' ' | tr '\n' ' ')
```
jprvita, line 53: This check does not work if the existing filesystem label has spaces on it. But you don't need to look into `/proc/mounts` at all, you can unmount with udisksctl, as you do on line 106 here.

jprvita: Also, you need to make sure you are unmounting all partitions of the target disk.

mschiu77: Actually this is OK for unmouting multiple partitions. I'll put the `/dev/sda1 /dev/sda2...` etc in one line by`tr '\n' ' '` and `umount` can unmount multiple partitions. But there's problem if the one of the device is busy. So I replace the `umount` with the `udiskctrl`.  But I still use `/proc/mounts` for parsing because the output of `lsblk` and `udiskctl` is not that easy to parse. So I keep it as-as.

PR #647, opened 2020-09-16
eos-core-depends: Replace gvfs-bin with libglib2.0-bin

pwithnall:
gvfs-bin has been removed in favour of libglib2.0-bin. The latter
contains different tools (the `gio` tool rather than `gvfs-copy` and
`gvfs-rename` etc.) but they are at least maintained.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

https://phabricator.endlessm.com/T30813

PR #648, opened 2020-09-16
eos-dev-suggests: Drop gnome-tweak-tool

andrunko:
Our current version is very old and doesn't work with latest GNOME and is causing some build failures on jenkins PR jobs.

The package got renamed to gnome-tweaks and latest from debian is 3.34.0. Let's remove it for now to fix the jenkins PR jobs and we can re-add later if needed.

### DIFF
--- a/eos-dev-suggests
+++ b/eos-dev-suggests
@@ -41,7 +41,6 @@ d-feet
 eos-dev-kernel
 evemu-tools
 gcovr
-gnome-tweak-tool
 gparted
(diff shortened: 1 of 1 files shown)


andrunko:
The gvfs-bin failure on CI is expected and actually what we are trying to fix - see https://github.com/endlessm/eos-meta/pull/647.

GeorgesStavracas (approved):

PR #649, opened 2020-09-16
eos-dev-suggests: Drop d-feet

andrunko:
This is causing a build failure on jenkins PR jobs and the app is available on flathub so let's drop it here.


andrunko:
As with https://github.com/endlessm/eos-meta/pull/648, the gvfs-bin failure on CI is expected and actually what we are trying to fix - see #647.

PR #650, opened 2020-09-17
eos-write-live-image: Reduce exFAT cluster size

wjt:
By default, mkfs.exfat picks the cluster size (allocation unit) based on
the disk size.  On large disks this is 128 KiB, wasting a lot of space on
Endless Key with many small files.  Force it to 8 sectors, ie 4096 bytes.

A crap program I wrote suggests that more than 5 GiB of space is lost by using a 128 KiB cluster size rather than 4 KiB on the Endless Key. There are something like 45,000 files less than 128 KiB on the disk and 20,000 less than 4 KiB so it might actually be true.

https://phabricator.endlessm.com/T30817


jprvita (approved):
Looks good, thanks!

PR #652, opened 2020-09-23
core: Add libfsapfs-utils package

dnohales:
This is to be able to access Apple DMG packages, necessary to create the
Endless Key image with Apple software.

https://phabricator.endlessm.com/T30829

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -156,6 +156,8 @@ imagemagick
 # Needed to support some Epson scanners
 imagescan-driver
 libblockdev-crypto2
+# Needed to unpack Apple DMG files for the Endless Key project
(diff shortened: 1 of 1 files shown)


eos-core-depends
```diff
@@ -156,6 +156,8 @@ imagemagick
 # Needed to support some Epson scanners
 imagescan-driver
 libblockdev-crypto2
+# Need to support accessing Apple DMG files for the Endless Key project
```
wjt, line 159:
```suggestion
# Needed to unpack Apple DMG files for the Endless Key project
```

wjt (approved):
Minor comment tweak!

PR #653, opened 2020-09-29
eos-core-depends: Add gnome-remote-desktop

andrunko:
This is needed for the GNOME screen sharing support.

https://phabricator.endlessm.com/T30882

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -122,6 +122,7 @@ gnome-keyring-pkcs11
 gnome-logs
 gnome-online-accounts
 gnome-orca
+gnome-remote-desktop
 gnome-screenshot
 gnome-shell-extension-appindicator
 gnome-software


eos-core-depends
```diff
@@ -122,6 +122,7 @@ gnome-keyring-pkcs11
 gnome-logs
 gnome-online-accounts
 gnome-orca
+gnome-remote-desktop
```
wjt, line 125: Hi, @jprvita! I recently claimed that GNOME screensharing works in Endless.  I'm sure it once did because I've used it, but evidently it has stopped at some point! I think you said some dependencies were deliberately excluded. This package will bring in `libvncserver1` (all its other dependencies are already installed). OK?

andrunko: Fwiw, this adds ~700kb to the ostree:
```
After this operation, 681 kB of additional disk space will be used.
```

PR #654, opened 2020-09-29
Add busybox for PAYG initramfs images

ManMower:
Add busybox for PAYG initramfs.

https://phabricator.endlessm.com/T30878


ManMower:
forced push changes comments that indicates busybox was only for PAYG initramfs, since we're going to use it in all initramfs.

Also updated the commit log for that patch similarly.

PR #656, opened 2020-10-09
eos-core: Install git

wjt:
Previously, git was installed only as a dependency of flapjack, which we
removed in 3adf120525d0e538b98e1dca0b9bf4ca90046001.

I believe it is too useful to exclude. However, it does add 38 MB to
the ostree.

https://phabricator.endlessm.com/T30935


andrunko (approved):
LGTM, we already had it before so let's not regress.

PR #660, opened 2020-11-12
Remove chromium in favor of flathub version

andrunko:
https://phabricator.endlessm.com/T30997


andrunko:
Converted to draft as I think we should only merge this and related changes once chromium hits flathub stable (atm we don't configure a flathub-beta remote and thus can't install it).

Appreciate any comments in the meantime.

eos-tech-support/eos-diagnostics
```diff
@@ -435,33 +435,6 @@ function collectTemperatureInfo() {
...
-    let widevinePath = Gio.file_new_for_path('/var/lib/chromium-plugins-extra/eos-chrome-plugin-updater/libwidevinecdm.so');
-    if (widevinePath.query_exists(null)) {
-        widevineOutput += tryReadFile('/var/lib/chromium-plugins-extra/eos-chrome-plugin-updater/Widevine_VERSION.txt',
-                                      'not available') + '\n';
-    } else {
-        widevineOutput += 'not installed\n';
-    }
-
-    return flashOutput + widevineOutput;
-}
```
wjt, line 463: Would we not want to adjust the Widevine check for the new path?

Or is the assumption that we can just use the version of the Widevine Flatpak runtime and/or assume Chrome has downloaded the right thing?

andrunko: That is a good point - I'm discussing with Ryan about this as we may end up not having a system-wide version of widevine in the end (as chromium flatpak already downloads it automatically into the user's folder). If we end up having a system-wide version it would make sense to keep this but not so much if local for the user.

andrunko: With latest flatpak the widevine cdm is handled/downloaded by chromium itself and has a per-user installation so let's drop this.

andrunko:
Updated to also include removal of `eos-chrome-plugin-updater` now that the widevine cdm is handled by the flatpak itself.

wjt (approved):

PR #662, opened 2020-12-02
eos-diagnostics: Add some more information to the output

ManMower:
For some issues, debugging is assisted by having a list of EFI variables
and their contents.

Additionally, let's dump the secure boot key and some rudimentary PAYG
information.

https://phabricator.endlessm.com/T31142

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -692,6 +692,52 @@ let diagnostics = [
         },
     },
     {
+        title: 'EFI variable list',
+        content: function() { return trySpawn('ls -l /sys/firmware/efi/efivars'); },
+    },
+    {
+        title: 'Secure Boot Platform Key',
+        content: function() { return extractCertificates(); }
+    },
+    {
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-diagnostics
```diff
@@ -918,6 +984,13 @@ ARGV.forEach(function(arg) {
         filename = arg;
 });
 
+let image_version = trySpawn('attr -q -g eos-image-version /sysroot') +
+                    trySpawn('attr -q -g eos-image-version /');
+
+if (image_version.startsWith("payg")) {
+    paygFlag = true;
+}
+
```
jprvita, line 993: Some payg image filenames do not start with "payg", but have "payg" as part of the personality string.

That said, we don't need to try to identify a payg image, we can try to fetch this information on all systems.

jprvita (changes requested):

ManMower:
Forced push rebases, and removes the check for payg image

jprvita (approved):

PR #665, opened 2021-01-12
eos-stage-*: Use CDN for dev repo URLs

dbnicholson:
With our shiny new Fastly CDN we can appropriately pass through HTTP
authentication to our origin server and don't need to access it
directly.

https://phabricator.endlessm.com/T31157


wjt (approved):
Nice!

PR #679, opened 2021-03-11
Added rsync as a os dependency

egrath:
Discussed at https://community.endlessos.com/t/packages-missing-in-development-version/16280


wjt:
Thanks!

PR #680, opened 2021-03-14
Support for gamer mouse configuration

LeandroStanger:
Discussed at https://community.endlessos.com/t/add-support-to-libratbag/16392?u=leandrostanger and https://community.endlessos.com/t/add-support-to-openrazer/16391?u=leandrostanger

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -230,3 +230,6 @@ xdg-desktop-portal-gtk
 yelp
 # Used by mutter for showing dialogs
 zenity
+# Support for configuring gaming mouse
+libratbag
+openrazer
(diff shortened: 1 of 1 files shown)


dsd:
Thanks for the contribution here, Leandro.
Please correct me if wrong but these packages seem very specific to gaming hardware which is distant from the focus of Endless OS Foundation & Endless OS. As such I would rather not have these present in our ostree packages.

I don't know enough about what these packages do to be sure if this is viable or not, but I would encourage exploration of how these packages could be made available within the flatpak runtime itself, perhaps relying on Flatpak portals for hardware access, so that these packages do not need to be available in the base OS.

PR #681, opened 2021-03-17
eos-download-image: Don't require --size with --url

wjt:
This was added to work around a CloudFront limitation: you cannot download more than 20 GiB in one go, nor can you even issue a HEAD request to find out the file size.

Fastly (and for that matter Bunny) does not have this limitation, so we can remove this artificial restriction.

(We could go further and remove all the logic that downloads files in 20 GiB chunks, but that code is harmless.)

This PR also trims out an old special-case that I am 99% sure is not used any more.

https://phabricator.endlessm.com/T23134

### DIFF
--- a/eos-tech-support/eos-download-image
+++ b/eos-tech-support/eos-download-image
@@ -286,9 +286,6 @@ class EosDownloadImage(object):
 
         self.args = args = p.parse_args()
 
-        if args.product == 'fnde' and args.personality == 'base':
-            args.personality = 'fnde_aluno'
-
         if args.url is not None and args.url.startswith(INTERNAL_BASE_URL):
             args.internal = True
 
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-download-image
```diff
@@ -21,6 +21,8 @@ KEYRING_SYSTEM_PATH = "/usr/share/keyrings/eos-image-keyring.gpg"
 
 # CloudFront limits the maximum file size to 20 GB
 # http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-limits.html
+# Our current CDN, Fastly, does not have this limitation, but fragmenting the
+# download is harmless.
```
dbnicholson, line 25: Right, I forgot about this part.

eos-tech-support/eos-download-image
```diff
@@ -258,7 +260,8 @@ class EosDownloadImage(object):
                        action='store_true',
                        help='Fetch images from the Endless internal network')
         p.add_argument('-s', '--size', type=int,
-                       help='Size in bytes of compressed image (use with --url)')
+                       help='Expected size in bytes of compressed image (for '
+                            'use with --url)')
```
dbnicholson, line 264: I kinda feel like all of this could go away since the size from the HEAD request is preferred, anyways. But I guess that makes it a bigger change.

dbnicholson (approved):

PR #682, opened 2021-03-18
Added support for L2TP and Cisco AnyConnect VPNs

egrath:
As suggested in https://community.endlessos.com/t/discussion-plugins-for-more-vpn-providers/16384


dsd:
add to whitelist

PR #683, opened 2021-03-23
Add important priority packages that we actually want so we can

BryanQuigley:
drop the others.

Backstory: https://community.endlessos.com/t/xterm-shows-in-3-10/16322/11

### DIFF
--- a/os-depends
+++ b/os-depends
@@ -10,6 +10,7 @@ dbus-user-session
 dbus-x11
 dconf-cli
 dirmngr
+dmidecode
 dosfstools
 dracut
 eject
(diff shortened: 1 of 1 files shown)


dbnicholson:
ok to test

dbnicholson:
Based on your analysis, I think you should add dmidecode, procps, systemd and fdisk. In the case of systemd, I think it's an oversight that it's not in there already. Even though it will come in via systemd-sysv, I'd rather explicitly say systemd is there in case debian wants to shuffle around the packages. The rest are utilities I'd want to ensure were available, and I'd rather be explicit about it than implicit.

BryanQuigley:
Sure, Added.

dbnicholson (approved):
Looks good to me, thanks!

PR #685, opened 2021-04-28
eos-core: Install tmate

wjt:
This tool is very useful for remote technical support. Having it
pre-installed in the OS simplifies the instructions we need to give to
the user who needs support.

https://phabricator.endlessm.com/T31863


wjt:
@dsd tagging you since we discussed this verbally

PR #687, opened 2021-05-03
eos-core-amd64: Drop printer-drivers-cnijfilter{2} & printers-driver-escp

starnight:
EOS is going to use driverless printers, so remove them.

https://phabricator.endlessm.com/T31859

### DIFF
--- a/eos-core-amd64-depends
+++ b/eos-core-amd64-depends
@@ -1,4 +1,1 @@
 eos-core
-printer-driver-cnijfilter
-printer-driver-cnijfilter2
-printer-driver-escp


eos-core-amd64-depends
```diff
@@ -1,4 +1,2 @@
 eos-core
-printer-driver-cnijfilter
-printer-driver-cnijfilter2
 printer-driver-escp
```
wjt, line 2: What about this one?

wjt: (I don't know anything about it.)

starnight: **printer-driver-escp** comes from [epson-inkjet-printer](https://github.com/endlessm/epson-inkjet-printer/blob/b36bfa80494401c244d633543da4185fb82bd7f3/debian/control#L164).

starnight: Search **epson-inkjet-printer** rpm (according to ["Initial import: files from the src.rpm file for the L100/L200 driver"](https://github.com/endlessm/epson-inkjet-printer/commit/1715d4501058862f96e3e9da0041741672dcf04a)) related around and only find [epson-inkjet-printer-escpr](https://rpmfind.net/linux/rpm2html/search.php?query=epson-inkjet-printer&submit=Search+...&system=&arch=).

starnight: Actually, Debian has **printer-driver-escpr**, which is on the dependent list of [printer-driver-all-enforce](https://packages.debian.org/bullseye/printer-driver-all-enforce).

dbnicholson (approved):
Yes, please!

PR #688, opened 2021-05-04
eos-core: Replace hplip with printer-driver-hpcups

starnight:
Follow Debian to strip all of the non-free hooks. Also do not ship hplip
at all, just the hpcups filters.

https://phabricator.endlessm.com/T31860

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -137,7 +137,6 @@ gsfonts
 gstreamer1.0-gl
 gvfs-backends
 gvfs-fuse
-hplip
 htop
 ibus-anthy
 ibus-avro


eos-core-depends
```diff
@@ -196,6 +195,7 @@ podman-toolbox
 # For modem support
 ppp
 printer-driver-all-enforce
+printer-driver-hpcups
```
ramcq, line 198: I don't think we need to add this, the printer driver all package above should depend on it.

ramcq (changes requested):

ramcq:
We still need to remove hplip from the metapackage I think?

starnight:
> We still need to remove hplip from the metapackage I think?
Yes!  I was distracted ...
Now, have the new commit!

PR #692, opened 2021-05-14
Remove vinagre from eos-core

wjt:
This app is not widely used on Endless OS. It is unmaintained upstream,
and there are more modern alternatives covering the same (or more)
functionality, such as Connections and Remmina, available on Flathub.

https://phabricator.endlessm.com/T31658


GeorgesStavracas (approved):
LGTM

PR #705, opened 2021-09-14
Remove all references to armhf

jprvita:
We don't support this architecture anymore after the EOS 3.8 series.

https://phabricator.endlessm.com/T30303


wjt (approved):
Oops, missed these. Thanks!

PR #706, opened 2021-09-15
eos-core: Drop htop

starnight:
EOS aleady has GNOME System Monitor, so drop htop.

https://phabricator.endlessm.com/T31569


starnight:
htop is an useful debugging tool.  Just going to hide the desktop icon with listing it in gnome-menus.

PR #708, opened 2021-09-24
Switch eog and evince to flatpaks

BryanQuigley:
Thread: https://community.endlessos.com/t/apps-that-have-flatpaks/17442/5

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -46,7 +46,6 @@ e2fsprogs-l10n
 efitools [amd64]
 # For flatpak-builder
 elfutils
-eog
 eos-acknowledgements
 eos-adblock-plus-extensions
 eos-b43fw-install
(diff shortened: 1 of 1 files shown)


wjt:
Could you split this and https://github.com/endlessm/eos-application-tools/pull/64 into one PR per app please?

My immediate concern is that GTK's print dialog uses Evince as its print previewer. I have confirmed that removing the `evince` package breaks print preview.

wjt:
Hmm, I noticed that the EOG Flatpak contains an H.265 decoder (for HEIC support), just as does GIMP. https://github.com/flathub/org.gimp.GIMP/issues/111 

GIMP is simply not preinstalled in contexts where that is a problem but I would feel more reluctant to remove the built-in image viewer…

BryanQuigley:
In my tests with gedit from flathub:
1) Run Gedit from flathub
2) Print Preview launches separate evince preview window.

1) Run gedit from deb
2) Print preview appears built-in.

How exactly did you test with the evince deb package removed?

wjt:
sudo eos-dev-unlock
sudo apt remove evince

Reboot to revert to the unmodified OS.

BryanQuigley:
Thanks for the hints. I think I'm going to try shotwell first.

PR #709, opened 2021-10-04
Update apt repo warnings

wjt:
Previously, our apt repos were not public, and our various un-ostree-ifying scripts warned about this.

Nowadays, they are public, but these tools are still discouraged. Update their warnings, and remove the long-obsolete eos-dev-fix.

https://phabricator.endlessm.com/T32540


jprvita (approved):
Looks great!

PR #710, opened 2021-10-08
tech-support: Make eos-stage-ostree stay on the same ref

jprvita:
This systems switching from prod to beta or beta to prod should stay on
the same ref they are already following (eoslatest1 for non-LTS, eosX
for LTS).

Systems switching from prod or beta to master will switch to following
the eosX.Y ref, and we don't know whether they were LTS or not before
the switch, so let's default to non-LTS.

https://phabricator.endlessm.com/T32639

### DIFF
--- a/eos-tech-support/eos-stage-ostree
+++ b/eos-tech-support/eos-stage-ostree
@@ -32,26 +32,26 @@ if [ $EUID != 0 ] ; then
     exit 1
 fi
 
-# Get the OS product.major version to build the branch series from. By
-# default, the system updates from the product version (e.g., eos3).
-major_version=$(. /etc/os-release && echo $VERSION | cut -d. -f1-2)
-product_version=${major_version%%.*}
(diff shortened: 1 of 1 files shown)


dbnicholson:
This is interesting and has interactions with https://phabricator.endlessm.com/T32628. Currently the reason that the ref gets mangled when you switch to the dev repo is because we can't be sure what the `eos3a` ref is if we're building 2 different eos3 branches at the same time. For example, if you build an eos3.8 ostree right now, it will overwrite the `eos3a` ref in the dev repo even though we clearly intend that to be an alias for eos3.9.

So, I think if we fix https://phabricator.endlessm.com/T32628 then this can be much simpler - just don't touch the deployed ref at all and assume the refs on the server are sane. Because what you have here has to make some assumptions about what the desired configuration is. What if this is an LTS build? Then we don't want to switch it to use `latest1`.

PR #719, opened 2021-12-28
core: Move gnome-terminal to Flatpak

starnight:
https://phabricator.endlessm.com/T32904


starnight:
Sorry!  Wrong package

PR #726, opened 2022-01-24
Remove flatpak-builder

wjt:
We take flatpak-builder directly from Debian, but maintain our own
Flatpak. Due to a security issue, flatpak-builder now has a versioned
dependency on a newer version of Flatpak than we have in the distro.

Remove flatpak-builder. We can readd it once we have updated Flatpak. In
the meantime the org.flatpak.Builder package from Flathub can be used
instead.

https://phabricator.endlessm.com/T32961

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -77,7 +77,6 @@ evolution-plugin-pstimport
 evolution-plugins
 file-roller
 flatpak
-flatpak-builder
 fonts-arabeyes
 fonts-crosextra-caladea
 fonts-crosextra-carlito


GeorgesStavracas (approved):
I don't know whether this change has deeper implications in our build stack or not, but I think this is a good idea. I'd like to have another opinion here before merging.

jprvita (approved):

jprvita:
> I don't know whether this change has deeper implications in our build stack or not, but I think this is a good idea. I'd like to have another opinion here before merging.

From https://phabricator.endlessm.com/T22335#922710 I don't think it has any implication on our build infrastructure. I'm merging this on the basis that this is the easiest path to get master builds going again, and we can discuss separately whether we should re-add it an how to do it.

PR #730, opened 2022-02-01
core: install orca, not gnome-orca

wjt:
Since 3.26.0, gnome-orca has been a transitional package which just
depends on orca.

https://phabricator.endlessm.com/T32987


wjt:
While I was here I searched the output of `dpkg -l` for `transition` and fixed a couple of others.

starnight (approved):

PR #735, opened 2022-02-15
Move PipeWire packages from os-depends to base-depends

GeorgesStavracas:
We need to install it at the same time libcanberra-pulse is
installed, otherwise it inadvertedly pulls the pulseaudio
into the image.

https://phabricator.endlessm.com/T33042

### DIFF
--- a/base-depends
+++ b/base-depends
@@ -13,6 +13,9 @@ gstreamer1.0-plugins-good
 gstreamer1.0-tools
 gstreamer1.0-vaapi
 ibus-gtk3
+# pipewire-pulse must be listed before libcanberra-pulse, otherwise
+# it'll pull pulseaudio
+pipewire-pulse
 libcanberra-gtk3-module
 libcanberra-pulse
 libcaribou-gtk3-module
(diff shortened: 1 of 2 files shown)


base-depends
```diff
@@ -13,6 +13,13 @@ gstreamer1.0-plugins-good
 gstreamer1.0-tools
 gstreamer1.0-vaapi
 ibus-gtk3
+# PipeWire must be listed before libcanberra-pulse, otherwise
+# it'll pull PulseAudio
+libspa-0.2-bluetooth
+pipewire
+pipewire-audio-client-libraries
+pipewire-pulse
+wireplumber
```
dbnicholson, line 22: I think just `pipewire-pulse` needs to move here. Apt can still order the installation correctly if all the other ones come later. All we're trying to do is help it figure out that we prefer `pipewire-pulse` to `pulseaudio` at the time it encounters the optional dependency in `libcanberra-pulse`. I think it's better to leave the other ones where they were to maintain the otherwise alphabetical ordering.

dbnicholson (changes requested):

base-depends
GeorgesStavracas: Sure

dbnicholson:
retest this please

dbnicholson (approved):

PR #738, opened 2022-02-24
Give our mimeapps.list priority over (Debian) GNOME's

wjt:
Currently, we install /usr/share/applications/mimeapps.list. Debian's
gnome-session package, which we have minor modifications to, installs
/usr/share/applications/gnome-mimeapps.list.

The [XDG mime-apps-spec states][0] that
$XDG_DATA_DIRS/applications/$desktop-mimeapps.list takes precedence over
$XDG_DATA_DIRS/applications/mimeapps.list, for each lowercased $desktop
from the colon-separated environment variable $XDG_CURRENT_DESKTOP (in
order).

On Endless OS today, XDG_CURRENT_DESKTOP is set to 'endless:GNOME'.

In some past Endless OS branches, we had patched the gnome-session
package not to install gnome-mimeapps.list, but some time later than
3.6.0 and no later than 4.0.0 we lost this change.

I think it is preferable to avoid patching the Debian package, and we
can use the existing XDG_CURRENT_DESKTOP mechanism to override it
without doing that. Rename the file to endless-mimeapps.list so that it
takes precedence over debian-mimeapps.list in the same directory.

https://phabricator.endlessm.com/T33210

[0]: https://specifications.freedesktop.org/mime-apps-spec/latest/ar01s02.html


wjt:
I haven't even built this package.

dbnicholson (approved):

PR #743, opened 2022-04-21
core: Add make

dbnicholson:
`make` is a generally useful utility that has no other dependencies
besides `libc6` and is only 1592 kB per the package `Installed-Size`. We
tend to think of `make` as a development tool, but people put all sorts
of recipes in a `Makefile`.

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -176,6 +176,7 @@ libpipewire-0.3-common
 libsasl2-modules
 # Parental controls
 malcontent-gui
+make
 # For running the image builder from Endless
 mmdebstrap
 modemmanager


dbnicholson:
@manuq was saying one of the things he missed on EOS was not having `make`. Several of the Kolibri build flows accommodate containers, but they're driven by a `Makefile`.

manuq:
Oh yes indeed, this is the only thing that I miss as a developer on Endless OS. And as far as I know `make` can't be installed at user level. From my experience, calling `toolbox run make` doesn't always work. So depending on the Makefile I may need to do: `toolbox run make --dry-run` and then copy/paste the output to the shell. And even that sometimes doesn't work if the Makefile is complex enough.

@wjt I wonder if you have a different experience as an ostree user, or a better way to workaround this.

dbnicholson:
I don't feel strongly about it. I'm sure there's a toolbox compatible workflow for it.

dbnicholson:
While potentially still useful, I think we can say for now that using a container is the preferred way to go here.

PR #744, opened 2022-04-26
eos-write-live-image: Get boot.zip etc from inside ISO

wjt:
In 7b48fb6057afc172cba3caf0d511ae258839427f "eos-write-live-image: Use
squashfs image from ISOs" we taught this tool to copy the
'endless.squash' file and its signature from within an .iso. However we
still required the .boot.zip, .boot.zip.asc and .img.asc files to be
alongside the .iso file.

These files can be found within the .iso, and with a relatively small
change we can use them directly.

### DIFF
--- a/eos-tech-support/eos-write-live-image
+++ b/eos-tech-support/eos-write-live-image
@@ -367,20 +367,33 @@ fi
 
 check_exists "$OS_IMAGE" "image"
 
+# Below here we start mounting stuff, so register a cleanup function
+trap cleanup EXIT
+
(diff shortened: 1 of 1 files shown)


wjt:
Draft because while I have tested that the script succeeds when given an ISO, the result (on a loopback FS) doesn't boot on my manually-constructed QEMU invocation; and I have not tested the non-ISO case.

jprvita (approved):
This LGTM, but yes, let's make sure all different paths are tested and the result boots.

PR #751, opened 2022-09-19
Make base-files with epoch become conflict

starnight:
EOS 5+ tries to install the base-files without epoch in the version string which uses the same schema as Debian's base-files. Therfore, add the prevoius EOS base-files with epoch into the conflict list.

https://phabricator.endlessm.com/T33521


dbnicholson (approved):
Thanks, that helps (a bit).

PR #755, opened 2022-10-25
os-depends: Install systemd-oomd

jprvita:
We are replacing our own psi-monitor with systemd-oomd.

https://phabricator.endlessm.com/T23914


jprvita:
ci, test this please

PR #766, opened 2023-08-29
eos-core-depends: Remove eog

cassidyjames:
Loupe is the new GNOME core Image Viewer; we should install it as a Flatpak instead of eog from the Debian repos; see https://github.com/endlessm/eos-image-builder/pull/157.

https://phabricator.endlessm.com/T34908


wjt (changes requested):
Please update https://github.com/endlessm/eos-meta/blob/master/debian/endless-mimeapps.list.

wjt (approved):
LGTM. I would bet that this list has not been updated in 10 years so there are probably some newer mimetypes like "webp" which we should add – but that's orthogonal to this change.

PR #767, opened 2023-11-03
Remove Brasero from the OS

wjt:
Brasero is unmaintained upstream and has no Flatpak version. It is very
lightly used – our metrics show just 44 users in September. Removing it
from the OS reduces our ongoing maintanence and may in due course allow
us to remove more of the multimedia stack from the OS.

There is at least one (proprietary) DVD-burner application on Flathub,
PowerISO, which we can direct users to as an alternative

https://phabricator.endlessm.com/T35003


starnight (approved):
Thanks!

PR #769, opened 2023-11-07
Increase test timeout time

wjt:
This test constructs a massive `apt-get install` line to check that the
metapackages are installable. It keeps hitting Meson's default 30-second
timeout in Jenkins.

Make the number bigger!


starnight (approved):
Thanks!  It helps!

This builds successfully on my laptop.  So, I'm going to merge it.

PR #772, opened 2023-11-20
eos-write-installer: Add `partprobe` to dependencies

cassidyjames:
Ran into this when using Silverblue; I needed to install `parted` to be able to use `partprobe`, otherwise it failed with:

```
eos-write-image: line 222: partprobe: command not found
```


dbnicholson (approved):
Makes sense. CI is broken right now for master, but this is certainly the right thing to do.

PR #774, opened 2023-11-27
Replace exfat-utils with exfatprogs

wjt:
exfat-utils came from the FUSE implementation of exFAT and have been
removed from Bookworm.

exfatprogs are from the team that submitted & maintain the in-kernel
exFAT implementation.

https://phabricator.endlessm.com/T35035


dbnicholson:
I think the CI is going to be fixed soon, but I'm going to merge this to keep things moving.

PR #779, opened 2024-01-11
Remove obsolete conflicts

wjt:
Spotted while investigating why we have three copies of WebKitGTK in the OS.


starnight (approved):
Thanks!

PR #781, opened 2024-01-12
eos-core-depends: Add console-common

starnight:
Install console-common to have the install-keymap utility, which is the recommended tool to specify a boot-time keymap to the system.

https://phabricator.endlessm.com/T35149


wjt (commented):
Nope, `plymouth` conflicts with `console-common`.

PR #783, opened 2024-01-19
eos-diagnostics: Collect extra PAYG logs

wjt:
eos-paygd writes extra logs to /var/log/eos-payg. These are useful on
systems where the journal is not persistent; and when we want deeper
history of PAYG state changes after the journal has been rotated.

I tested the following error cases:

- /var/log/eos-payg does not exist or is not readable; in this case the
  section is omitted. This should not happen in practice – the
  tmpfiles.d snippet sets mode 755.
- A file within is not readable; in this case the exception is included
  in the diagnostic file.
- A file within contains invalid UTF-8; in this case the replacement
  character is used.

If the directory is empty, the function returns empty string, which
means the section is omitted.

https://phabricator.endlessm.com/T33896

### DIFF
--- a/eos-tech-support/eos-diagnostics
+++ b/eos-tech-support/eos-diagnostics
@@ -763,6 +763,33 @@ let diagnostics = [
         }
     },
     {
+        title: 'PAYG logs',
+        content() {
+            let logDir = Gio.File.new_for_path("/var/log/eos-payg");
+            let enumerator, fileInfos;
+
+            try {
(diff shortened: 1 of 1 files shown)


wjt:
I don't have a real PAYG system to hand to test this on, and I deleted the VM & image I downloaded & set up the other day in anger.

wjt:
Tested this in a VM. It works. Self-merging.

PR #785, opened 2024-03-05
Remove gstreamer1.0-libav from OS

dbnicholson:
gstreamer1.0-libav provides a gstreamer plugin that proxies to ffmpeg's codecs via its various libav* libraries. In theory that's nice since gstreamer applications can gain access to ffmpeg's codecs. In practice we were trimming down ffmpeg so much that it probably does not provide codecs for any media types that weren't already provided by native gstreamer plugins.

Furthermore, nothing we ship appears to have a hard dependency on the gstreamer elements provided by the libva plugin. We've also dropped the `/var/lib/codecs` system that provided the less patent encumbered ffmpeg extra libraries.

By extension, this also removes the ffmpeg libraries since gstreamer1.0-libav was the only thing we shipped using them. That simplifies our development by not having to analyze the legality of all the ffmpeg codecs and adjust the build to fit.

https://phabricator.endlessm.com/T35225

### DIFF
--- a/base-depends
+++ b/base-depends
@@ -8,7 +8,6 @@ fonts-lato
 fonts-liberation
 fonts-wqy-microhei
 gnome-accessibility-themes
-gstreamer1.0-libav
 gstreamer1.0-plugins-good
 gstreamer1.0-tools
 ibus-gtk3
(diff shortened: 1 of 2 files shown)


wjt (approved):
Looks good. I wonder if we should add a `Conflicts` on some relevant package to avoid ffmpeg sneaking back in.

dbnicholson:
> Looks good. I wonder if we should add a `Conflicts` on some relevant package to avoid ffmpeg sneaking back in.

I definitely thought about it, but I was worried that would be painful for container users where you can install packages at will.

PR #787, opened 2024-03-07
mimeapps: Remove endlessm-app: handler

wjt:
eos-launch.desktop has been removed, and the use of the endlessm-app: scheme with it.

https://phabricator.endlessm.com/T31635


manuq:
Ah I see this was a missing bit of the URL scheme removal. Great.

PR #792, opened 2024-03-26
Mimeapps: Replace Rhythmbox with Decibels

cassidyjames:
Some of these look like old mimetypes, but for simplicity, replace all Rhythmbox ones with Decibels, plus add new Decibels mimes.

Depends on:

- [x] https://github.com/endlessm/eos-image-builder/pull/158

https://phabricator.endlessm.com/T32696


cassidyjames:
Hm, revisiting this, it looks like either Decibels is missing types it can handle, or something else is weird here. Investigating…

wjt (approved):
OK to merge once we have updated the image builder's default apps.

PR #798, opened 2024-05-07
eos-download-image: Use packaging.version.Version

wjt:
distutils.version is deprecated:

    $ python3 -c 'import distutils.version; distutils.version.LooseVersion("1")'
    <string>:1: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.


dylanmccall (approved):
This works for me!

PR #799, opened 2024-05-07
eos-write-installer: Support images without .asc signature; fix finding image partition after repartitioning

wjt:
For 3 years eos-installer has supported installing images that do not have an `.asc` detached GPG signature, only a `.sha256` file containing a SHA256 digest of the image. This is to allow installing images built by third parties, who do not have access to our image signing key.

However, `eos-write-installer` has continued to insist on the `.asc` file being present.

Fix this; and fix a recently-introduced regression.

https://phabricator.endlessm.com/T35362

### DIFF
--- a/eos-tech-support/eos-write-installer
+++ b/eos-tech-support/eos-write-installer
@@ -21,7 +21,7 @@ if [ ! -f "$EOS_DOWNLOAD_IMAGE" ]; then
     EOS_DOWNLOAD_IMAGE='eos-download-image'
 fi
 
-ARGS=$(getopt -o i:o:luh -l "installer:,os-image:,latest,update-installer,help" -n "$0" -- "$@")
+ARGS=$(getopt -o i:o:luh -l "installer:,os-image:,latest,update-installer,debug,help" -n "$0" -- "$@")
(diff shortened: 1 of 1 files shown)


eos-tech-support/eos-write-installer
```diff
@@ -316,8 +316,8 @@ if [ -f "$INSTALLER_IMAGE" ]; then
     # At this point udev should have seen that sfdisk closed the block device
     # which was opened for writing, so it will be reprobing.
     # Add udevadm settle here to avoid races.
-    udevadm settle
```
dsd, line 299: I think there is a reason to leave this settle in this position (in addition to adding the one 2 lines down). I think in response to sfdisk closing the block device, udev will momentarily remove and then recreate the block device. That would race with partprobe trying to open it on the following line. The race is avoided with this settle, making sure udev has finished doing its thing.

wjt: Revised to do both, and update the comment(s).

dylanmccall (approved):
Oh, I'm glad you spotted that regression! This looks like a good change and it is working well for me.

PR #808, opened 2025-07-22
Install eos-tech-support and endless-mimeapps.list with meson

### LINKED ISSUE #35: Install tools of eos-tech-support and endless-mimeapps.list with meson
We are going to install tools of eos-tech-support into EOS 7 #34.  However, they are installed by deb tool now.  But, EOS 7 uses BuildStream as the build tool.  So, we should install the scripts and `endless-mimeapps.list` with meson, instead of deb tool.

starnight:
EOS 7 will be based on GNOME OS, which uses BuildStream as the build tool. So, install eos-tech-support's scripts and `endless-mimeapps.list` with meson, instead of deb tool. The meson works in both EOS 6 and 7's build environment.

Fixes: https://github.com/endlessm/eos-build-meta/issues/35


wjt:
This seems fine as an interim step.

Since 2021 we have had an open issue to split these tools out of `eos-meta`: https://phabricator.endlessm.com/T31742

But presumably in due course the actual `meta` side of this package will be removed - the Debian metapackages will not be used, `eos-build-meta` will replace it - so actually we can do the opposite: delete the meta stuff and rename the source package here.

As mentioned in that ticket there are lots of tools here we can drop. We should review the list. But that can also be a follow-up.

PR #809, opened 2025-11-05
os: Install linux-image-generic or linux-rpi on arm64 platforms

starnight:
Going to have a new OSTree branch for RPi device. And, the OSTree branch installs linux-rpi for having better hardware support on RPi devices. So, install either linux-image-generic, or linux-rpi on arm64 platforms. The OSTree build configuration decides installing which linux kernel.

https://app.asana.com/1/1203676861188277/project/1211061530470838/task/1211622697341771


starnight:
Wrong target branch.  It should be eos6.0.

PR #813, opened 2026-04-29
eos-core-depends: add libmbim-utils

ramcq:
ModemManager needs mbimcli to run fcc-unlock.d scripts for my Quectel WWAN modem

### DIFF
--- a/eos-core-depends
+++ b/eos-core-depends
@@ -162,6 +162,8 @@ libgphoto2-l10n
 libgpod-common
 # Assuming useful for cups
 libmagickcore-6.q16-6-extra
+# Needed for ModemManager fcc-unlock scripts
+libmbim-utils
 # Assuming needed for MTP support
 libmtp-runtime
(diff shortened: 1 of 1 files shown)


eos-core-depends
```diff
@@ -164,6 +164,8 @@ libgpod-common
 libmagickcore-6.q16-6-extra
 # Assuming needed for MTP support
 libmtp-runtime
+# Needed for ModemManager fcc-unlock scripts
+libmbim-utils
```
starnight, line 166: This is okay, but please move libmbim-utils upper than libmtp-runtime for the characters order.  It is good to be merged after the fix. :)

starnight (changes requested):

eos-core-depends
ramcq: oh yes... 😆

starnight:
Cherry picked the commit to eos6.0 branch.

